PR #5, opened 2016-12-02
English language updates

rocco27:
Missing one string (omano overtook me :))
Few text corrections


rocco27:
About changing text in Watch logger When, Who, Where, What: well it was simple and okey, but I think it is better to give names to things.

PR #6, opened 2016-12-04
Add localization to menu names for DSi module

Zorrototo:


rocco27:
I already missed those strings, good catch!

Zorrototo:
I also plan to localize the module too later. But so much things to do before.

PR #8, opened 2016-12-14
Support for installations behind Cloudflare and nginx>apache proxy

Adjokip:
REMOTE_ADDR is always going to be set, therefore the current check is
pointless when trying to support the other variables. Additionally,
users running an installation behind Cloudflare will only get the
Cloudflare server IP that the remote user is being proxied by. And users
such as myself who run an nginx -> apache2 proxy (apache2 serving the
dynamic content) will be getting the local IP address logged. This
change is more important for the logger() functionality and fixes the aforementioned issue.


own3mall:
Updated.  Thanks.

https://github.com/OpenGamePanel/OGP-Website/commit/4660cf488d9d89433855b302c2ed667e062813aa

PR #9, opened 2016-12-16
Fix GameTracker map images not being loaded.

Adjokip:
A fresh installation of OGP fails to get map images from GameTracker.
I'm not really sure why - but this fixes that and is a simpler check.

### DIFF
--- a/includes/database_mysql.php
+++ b/includes/database_mysql.php
@@ -2406,18 +2406,16 @@ class OGPDatabaseMySQL extends OGPDatabase
 	
 	public function logger($message){
 		$user_id = isset($_SESSION['user_id']) ? $_SESSION['user_id'] : 0;
-		if ( isset($_SERVER["REMOTE_ADDR"]) )
-		{
-			$client_ip = $_SERVER["REMOTE_ADDR"];
+		if(isset($_SERVER['HTTP_CF_CONNECTING_IP']) === true){
+			$client_ip = $_SERVER['HTTP_CF_CONNECTING_IP'];
+		}elseif(isset($_SERVER['HTTP_X_FORWARDED_FOR']) === true){
+			$client_ip = $_SERVER['HTTP_X_FORWARDED_FOR'];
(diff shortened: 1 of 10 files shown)


own3mall:
Please provide steps on how I can reproduce this issue.  I don't like the fixed substr approach either.

Adjokip:
It quite simply happens after a fresh install for me.

It's happened on my live server where I've hosted OGP for the past ~20 days or so. I decided to take a look at the code locally on a VM to see what causes this among other minor annoyances (seen in my last commits)

See the following image for an overview of the Game Monitor on a fresh install - and note that these are all default maps and are on GameTracker:
fireshot screen capture 009 - open game panel game manager - game monitor - 192_168_1_106_home_php_m gamemanager p game_monitor

I'm not really sure why that simple change 'fixes' it, the header check seems right - I ran var_dump(); on the get_headers(_curl) functions and the if comparison should be working right.

Adjokip:
Actually, nevermind the substr check. This seems to fix it:

`if(trim($file_headers[0]) == 'HTTP/1.0 200 OK' || trim($file_headers[0]) == 'HTTP/1.1 200 OK') return $path;`

rocco27:
I can confirm when I use new codess I get back map images for CS:GO, CoD4, Minecraft game servers.

own3mall:
Why did you rename get_start_cmd to get_sart_cmd?

Unless there is a good reason for that, you should really try to avoid making changes like that.

Adjokip:
The initial commit I made in #8 358acc6 corrected the typo from sart to start.

Why? Readability and clarity as to what the function does. However that exact change was never merged with the main branch according to the commits overview.

To keep my local copy updated before suggesting anything else, `git merge` copied what was on the master branch and therefore the rename was undone.

rocco27:
I understand it was an accident, but if you pull anything check what you send back to source...;)

own3mall:
Ah, I see.  Thanks for the clarification Adjokip.  I didn't realize it was named incorrectly to begin with.

PR #15, opened 2017-01-05
Better Killing Floor 2 support

DACRepair:
- Added LGSL config for KF2 
- Changed KF2 config, adding / changing the following:
  - Changed port mapping so that one could run several KF2 servers. Example:
    - Game Port: 7777 (Default: 7777)
    - Query Port: Game Port +1 (Default: 27015)
    - Web Admin Port: Game Port + 2 (Default: 8080)
- Did some general cleanup to keep the UT-Style args clean.

### DIFF
--- a/modules/config_games/server_configs/killingfloor2_win64.xml
+++ b/modules/config_games/server_configs/killingfloor2_win64.xml
@@ -1,19 +1,21 @@
-<game_config> 
+<game_config>
   <game_key>killingfloor2_win64</game_key>
   <protocol>lgsl</protocol>
-  <lgsl_query_name>redorchestra2</lgsl_query_name>
+  <lgsl_query_name>kf2</lgsl_query_name>
   <installer>steamcmd</installer>
   <game_name>Killing Floor 2</game_name>
   <server_exec_name>KFServer.exe</server_exec_name>
-  <cli_template>%MAP%%GAMEMODE%%DIFFICULTY%%GAMELENGTH%%PLAYERS% %PORT% %IP% %WEB_ADMIN_PORT%</cli_template>
+  <cli_template>%MAP%%GAMEMODE%%DIFFICULTY%%GAMELENGTH%%PLAYERS%%IP%%PORT%%QUERY_PORT%%WEB_ADMIN_PORT%</cli_template>
   <cli_params>
     <cli_param id="MAP" cli_string="" />
-    <cli_param id="IP" cli_string="-MultiHome=" />
-    <cli_param id="PORT" cli_string="-Port=" />
-	<cli_param id="PLAYERS" cli_string="?MaxPlayers=" />
+    <cli_param id="IP" cli_string="?MultiHome=" />
+    <cli_param id="PORT" cli_string="?Port=" />
+    <cli_param id="PLAYERS" cli_string="?MaxPlayers=" />
+    <cli_param id="CONTROL_PASSWORD" cli_string="?AdminPassword=" />
   </cli_params>
   <reserve_ports>
-    <port type="add" id="WEB_ADMIN_PORT" cli_string="-WebAdminPort=">10</port>
+    <port type="add" id="QUERY_PORT" cli_string="?QueryPort=">1</port>
+    <port type="add" id="WEB_ADMIN_PORT" cli_string="?WebAdminPort=">2</port>
   </reserve_ports>
   <map_list>maplist.txt</map_list>
   <console_log>KFGame/Logs/Launch.log</console_log>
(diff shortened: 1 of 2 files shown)


modules/config_games/server_configs/killingfloor2_win64.xml
```diff
@@ -49,27 +51,27 @@
     <param id="GAMEMODE" key="?Game=" type="text">
       <option>ns</option>
       <caption>Game Mode</caption>
-      <desc>Leave empty to run the normal game mode.&lt;br&gt;To run your server with official Versus Survival mode use this:&lt;br&gt;&lt;br&gt;KFGameContent.KFGameInfo_VersusSurvival&lt;br&gt;&lt;br&gt;You can use any other gametype if you installed any other mod from Steam Workshop</desc>
+      <desc>Leave empty to run the normal game mode.&lt;br&gt;To run your server with official Versus Survival mode use this:&lt;br&gt;&lt;br&gt;KFGameContent.KFGameInfo_VersusSurvival&lt;br&gt;&lt;br&gt;You can use any other gametype i$
```
LargoUsagi: This is malformed XML and will not process.

DACRepair:
Nano messed up a line from my XML (Wordwrap, will be resubmitting)

DACRepair:
Fixed the word wrap issue.

own3mall:
Adding one for the query port would make 7777 7778, right?  So the def in lgsl should be?  7778 by default instead of 27015?

Zorrototo:
The XML seems OK, but you use URL ARGUMENTS instead of SWITCHES like explained in the official documentation http://wiki.tripwireinteractive.com/index.php?title=Dedicated_Server_(Killing_Floor_2)#Command_Line_Launch_Options so most of the parameter generated on launch are not like it 'should' be, but it still seems to work as intended (Port, QueryPort, WebAdminPort, and MultiHome are SWITCHES not url arguments). In my opinion better stick to how it is supposed to be, in order to not break anything after an update that will make your changes not work.

Also KF2 server automatically sets the Steam Port (default 20560, it is calculated from the Port so it adds to it 20560-7777=+12783), maybe we can add this to the Reserve Ports if it opens this port in the machine firewall when OGP manages the firewall? (I don't manage firewall with OGP so I'm not aware about how this works). Just my two cents here.

Also I'm not a fan of forcing the admin password through the start command line (I only use OGP as an admin, so I'm not sure if a simple user with basic rights can click EDIT in the server page and change the CONTROL PASSWORD from this page, when I'm sure he always can set it in his config file), but why do you use the CONTROL_PASSWORD from the server EDIT page, if you do not use it in your start command line?


I think some tweak can still be done now we're at it.

I have an idea too force the enabling of the Web Admin, with the Custom Fields, by modifying the DefaultWeb.ini this way the server will always have the Web Admin set to the defined parameter bEnable=TRUE instead of =FALSE (by default it is set to FALSE so owner has to manually enable it). I still did not test it but will submit this change later after tests.

I think we can hold on a little before validating these changes.

LargoUsagi:
@Zorrototo For the webadmin bEnabled, You could also just edit the KF version of the config by running a post install to copy the defaults to the KF-* version of the ini files before the server boots the first time.  This way the server still works as it would be documented and you are not getting rid of that every time you update the game.

Zorrototo:
No because of UE3/KF2 resetting config files sometimes for weird reasons. It needs to be generated from the DefaultWeb.ini file first. same kind of problems (resetting config) happens after enabling multi admin support, if you import an old MultiAdmin.ini file with all your admins infos, it will most likely reset and then you can edit this fresh almost empty config file to re-add your multiple admins infos, 

Editing the DefaultWeb.ini is a safer way to do it in my opinion to avoid unwanted reset, and in case of Day Saving time change and/or time zone modification on the machine (see the bug report here http://forums.tripwireinteractive.com/forum/killing-floor-2/technical-support-ae/the-bug-report-ae/116399-time-zone-change-resets-some-config-files for more info), it will still be enabled by default, because it is always modified at server start.

Still it was an idea I did not test yet and I'm not sure about everything.

own3mall:
These changes require more research, testing, and tweaks before they can be merged into master.  As far as I know, there were no problems with the original config file to begin with.

DACRepair:
Agreed, I will revisit some of the changes I did, and try to rework them. 
My main concern with the ports, was that 27015 is an overlapping game port with many other games (mainly source based). Granted you can change these, but it was more of an idea of containing each server into a "block" so that when you start just creating servers, you don't overlap. I could try working around keeping the port format more "default", I just see ports as more "real estate" than anything.

I'm gonna shoot for getting another commit in to update this today.

own3mall:
I'm going to close this for now.  If you fix the issues and concerns listed above, please open a new pull request.  Thanks.

PR #22, opened 2017-02-02
Make it possible to use the already present limits

Zorrototo:


own3mall:
Just curious:  Why did you float the tables left and right?

Zorrototo:
I wanted to have the "View XXX per page" part on the right and couldn't find a better way without destroying the tables. Feel free to modify it if you know a better way :)

own3mall:
I changed it a little.  Thanks for finding and fixing this issue!

PR #24, opened 2017-02-02
Avorion server (steam)

bimbam2103:
This is an updated version to use with the avorion server (steam).
Thanks to omano and rocco27 for the great help.


Zorrototo:
http://www.avorion.net/forum/index.php?topic=973.0 a current bug in Avorion makes it not reachable if the polling port is modified, safer to not set it for now (but will use always the 27003 port for now)

bimbam2103:
Zorrototo is correct. There is a bug in avorion server.

Please ignore the reverts if possible. The pull is correct. The server script works well (except the bug)

Zorrototo:
I sent a modification with GitHub Desktop it appears in my client, seems to be sent, but apparently not because I got 404 error when I click the GitHub icon on my pull request.. weird..

I removed "%POLLING_PORT%" from CLI and the corresponding reserved port. If you can remove it when accepting that will avoid problems for the one unaware of the bug.

own3mall:
Ok, please let us know when we can update it back :)

PR #25, opened 2017-02-03
Fixed the LD_LIBRARY_PATH need, removed %MAP%

Zorrototo:
Quick fix regarding this thread http://www.opengamepanel.org/forum/viewthread.php?thread_id=5186

Changed server executable to be the SH script I added as post_install.
Removed %MAP% from CLI as it was instantly starting, on a map called Bootstrap instead of the Procedural map, server was kinda 'working' (you can see yourself connecting/disconnecting from server in the logs), but you only join a black screen with a sleeper icon, and can't spawn in this actually non-map. Commented the MAP server parameter for now until I get more time to see why map is not selectable anymore (they might have removed that, as I don't even see a mention about that parameter in Steam official documentation).

### DIFF
--- a/modules/config_games/server_configs/rust_linux64.xml
+++ b/modules/config_games/server_configs/rust_linux64.xml
@@ -4,9 +4,9 @@
   <gameq_query_name>rust</gameq_query_name>
   <installer>steamcmd</installer>
   <game_name>Rust</game_name>
-  <server_exec_name>RustDedicated</server_exec_name>
+  <server_exec_name>RustDedicatedOGP.sh</server_exec_name>
   <query_port type="add">1</query_port>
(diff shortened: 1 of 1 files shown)


rocco27:
I will taka a look sometime.

rocco27:
So here is the answer %MAP% why doesn't work why does it load bootstarp map instead of choosen map:
If I choose Procedural Map then OGP shows  +server.level "Procedural Map" in the startup command line.
But if we check the console.log of Rust we can see the order/command sent load server.level: "Procedural" only, missing the space and the Map tag.

If I use Baren map, then it loads Baren map and works fine of course.

I already found another servers if the server name or host name has a space(s) and we use in the startup command line, then  the gameserver has shorted name.

It is same with another characters, if we use special characters like : # etc, the OGP cut the server name.

Zorrototo:
For the server name I see same problem with Avorion, if there is a space in it server name is only one word, with "s" or "sq" options.

own3mall:
Does the command line show the name correctly though just before the server is started in the panel?  This is a bug you're reporting?

Zorrototo:
As I told you on Steam the other day, I don't know what you did but it is now showing the server name properly with more than one word (after you added the environment variable in XML).

PR #26, opened 2017-02-03
Screenlog management on server deletion

Adjokip:
As per what @rocco27 suggested awhile ago, this adds some sort of control over server logs for servers that are about to be deleted ( See: http://www.opengamepanel.org/forum/viewthread.php?thread_id=3214#post_19798 )

It adds 3 button elements. Clicking on them dynamically updates the href value of the 'Yes and delete the files link' link - adding another GET variable for parsing.

It also includes what @Zorrototo suggested for the ability to backup the log files, perhaps for later inspection.

Tested on Ubuntu and Windows, works on both.

Example

### DIFF
--- a/modules/user_games/del_home.php
+++ b/modules/user_games/del_home.php
@@ -90,23 +90,22 @@ function exec_ogp_module() {
 		no . "</a></p>";
 		
 		// Not the prettiest way to do this...
-		echo <<<HTML
-		<script>
+		echo '<script>
 			$(function(){
 				var linkElement = $("#deleteLink"),
 				defaultLink = linkElement.attr("href"),
 				newLink = linkElement.attr("href");
 				
-				$('#doBackup').change(function(){
+				$("#doBackup").change(function(){
 					if($(this).is(":checked")){
-						linkElement.attr("href", newLink + '&logAction=backup');
+						linkElement.attr("href", newLink + "&logAction=backup");
 					}else{
 						linkElement.attr("href", defaultLink);
 					}
 				});
 			});
-		</script>
-HTML;
+		</script>';
+		
 		return;
 	}
 
--- a/modules/user_games/del_home.php
+++ b/modules/user_games/del_home.php
@@ -80,9 +80,7 @@ function exec_ogp_module() {
 			{
 				echo "<p><a href=\"?m=user_games&amp;p=del&amp;y=y&amp;home_id=$home_id&amp;files=y\" id=\"deleteLink\">" . 
 					 yes_and_delete_the_files . "</a> ";	
-				echo <<<HTML
-					<input type="checkbox" name="logAction" id="doBackup"><label for="doBackup">Delete and Backup Logs</label> |
-HTML;
+				echo '<input type="checkbox" name="logAction" id="doBackup"><label for="doBackup">Delete and Backup Logs</label> |';
 			}
 		}
 		else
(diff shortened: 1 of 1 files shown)


own3mall:
I'm not sure why this feature is needed.  Why wouldn't you just make a backup yourself if you wanted to keep the log files?  Or better yet, why would you remove the user's files in the first place even though you're deleting the server?  I usually leave all files alone, but that's me.  Not really sure cluttering the UI for such a minor feature is a good idea, but again, that's just my opinion.

Also, shouldn't there be a copy command added after the .tar.gz backup file is created to put them in a centralized directory such as in the OGP folder somewhere (under a sub folder)?  If you create a backup of the logs, and then delete the server including its files, wouldn't it remove your backup file too?

Adjokip:
Typically on all my nodes, I do have automated backups. The files aren't necessarily an issue - it was more to fulfil a request which I noticed could be easily done within a few minutes.

The log deletion is optional, either not clicking a button _or_ clicking 'do nothing' offers the default behaviour as-is currently.

It does not remove the backup which was just created, it removes files based on the `screenlogs/screenlog.OGP_HOME|UPDATE_000000000` prefix, and the `home_id_x` directory, whereas backups are stored as `screenlogs/home_log_backup_id.tar.gz`

If UI cluttering is an issue, it'd always be possible to 'Screenlog Handling' be clicked and then dropdown, revealing the buttons. But as I mentioned, this was only to fulfil a request which had been sitting around.

own3mall:
Gotcha.  I understand now.

own3mall:
Couldn't we just put a checkbox somewhere on the existing line that says "Backup Log Files Before Deleting"?  Delete Logs and Do Nothing actually don't do anything, right?  So couldn't we just add a checkbox to the end of the Yes Delete Files, Yes, No that does the same thing?  Just thinking...

own3mall:
More specifically, place that checkbox next to the "Yes and Delete the files" only option since I'm assuming that's the only operation that is destructive?

Adjokip:
Yeah - 'Do Nothing' was just a button to reset the uri so the logHandling() code wasn't executed... and, 'Delete Logs' only set the uri so the logs were deleted but not backed up.

I made a change so there's a checkbox instead, commiting it now.

own3mall:
Change it to a normal echo statement (use single quotes if you don't want to escape the double quotes), and I'll merge it.  I've never seen that type of echo before and can't seem to find documentation on it though I didn't try too hard.

```
echo <<<HTML <input type="checkbox" name="logAction" id="doBackup"><label for="doBackup">Delete and Backup Logs</label> | HTML;
```

Adjokip:
Intended to do that before commiting.

That echo is the HEREDOC synatax, it's good for large blocks that are being output or assigned. http://php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc

own3mall:
Please move your JS functions to document.ready in js/modules/user_games.js file, get rid of the HEREDOC syntax anywhere else you've used it, and then I'll merge :+1:

Adjokip:
`js/modules/user_games.js` doesn't seem to be included in `home.php?m=user_games&p=del` ..? I don't see it in the page source or the Chrome Network log - so I'm unable to do that.

The Heredoc echos have been removed now, via my last commit.

own3mall:
Yes, you're right.  OK, looks good.  Thanks!

PR #27, opened 2017-02-03
Bug fix avorion server

bimbam2103:
Due to an actual bug in avorion server polling_port and port reservation
was removed.


bimbam2103:
It was already corrected. So this pull request it not necessary.

PR #36, opened 2017-02-12
Wrong duplicate string

Zorrototo:


own3mall:
Asking rocco to fix this one at Transifex.

rocco27:
It is done. Ready to translate on Transifex. ;)

rocco27:
Readded my fix again. I was so blind before. I focused only to PureFTPd tag... :D

enbled -> enabled

PR #38, opened 2017-02-13
using native KF2 in LGSL, adding reserve ports, and new maps

Zorrototo:
As @DACRepair did not resend his pull request after the notes and suggestion I made, I do it myself.
One thing I'm not sure is the STEAM_PORT I added in the reserve port, I removed the cli_string as there is currently none, and as I'm not using OOGP to manage my firewall, I'm not sure if it will still reserve the port and open it in firewall. Feel free to remove this line 18 in case it is wrong without the cli_string


Zorrototo:
just to mention, these reserve ports are not randomly picked. they are the ports already in use when we do not set them (see a KF2 log it says it sets automatically the port from game port when they are not set). So it will use exactly same ports as it uses now without these reserve ports.

PR #39, opened 2017-02-14
Completed the german language in global.php

bimbam2103:


FarukGamer:
It would be great if you use Transifex for this.
https://www.transifex.com/opengamepanel/ogp/

Zorrototo:
Yep translations are maintained through Transifex. Register there, and ask to join the German team :) https://www.dropbox.com/s/aib722xtw5rlxcl/Capture%20d%27%C3%A9cran%202017-02-14%2023.57.44.png?dl=0

Maybe Rocco can upload your files to Transifex, I can't guarantee that you'll have to see with him, but if you plan to continue translation, stop doing it this way, and use the friendly interface at Transifex which is linked to GitHub repositories.

Sad you didn't see the thread about that at OGP forum http://www.opengamepanel.org/forum/viewthread.php?thread_id=5057

rocco27:
Hello there! I strongly recommend to start to use Transifex. We already re-organized every languages before we started to use Transifex. Now we sync language files from Transifex to GitHub only and not reverse way. The reason is in the previous sentence. If you give a try and translate your global.php (or just do copy&paste) for every strings, then it will be a nice practice in the first time and see how nicely it works.
You can join anytime to OGP Transifex Project, if you want you can be member of the translation team as well. Thanks for your understanding!

bimbam2103:
Hello everyone.

No.. :( I didn't see the post before. Sorry for that. I give it a try. Thank you

bimbam2103:
I closed the request because i completed already more files than by editing the source files :)

That's a great portal for translating. If i have enough time i will complete the german translation.

Greetings

Issue #19, opened 2017-01-17
Options parameter of server_params doesn't work in the XML files


rocco27:
It is a very-very old issue. I already reported to DieFeM before and he said he will check it, but he is inactive...

In the xml files we have cli_params, reserve_ports, replace_texts, server_params and custom_fields where we can use **options=""** for first two and **<options></options>** for the rest to allow or disallow to use space and queries or nothing beetwen the server parameters and values.

The options work everywhere instead of server_params. Did it ever worked? Is it a bug or?
Here is an for example:
```
  <server_params>
   <param key="?ModID=" type="text" id="MODID">
      <default></default>
      <caption></caption>
      <options></options>
      <desc>Set Mods.</desc>
    </param>
  </server_params>
```

Then I have this in the server startup command line:
TheCenter?listen?Multihome=XXX.XXX.XXX.XXX?Port=XXXX?QueryPort=XXXX?MaxPlayers=XXX?ServerAdminPassword=XXXXXXXX ?ModID= "1234567" -server -log

So when I use nothing I have space and quotes for MoDID parameter. It doesn't matter what I set, it is always forced to use a space and quotes!

own3mall:
So try it with a blank options element in your game server config XML:

`<options></options>`

The various possibilities are:

`if ($options == "tags")
				$file_content = preg_replace("/(<$default$var>)(.*)(<\/$default>)/m", '${1}'.$info_param.'${3}', $file_content, 1);
			elseif ($options == "tagValueByName")
				$file_content = preg_replace('/('.$default.'.*name="'.$var.'".*value=)(".*")/m', '${1}"' . str_replace('"', '\"', $info_param) . '"', $file_content, 1);
			elseif($options == "s")//separated
				$file_content = preg_replace("/$default/m", "$var $info_param", $file_content, 1);
			elseif ($options == "q")//quoted
				$file_content = preg_replace("/$default/m", "$var\"" . str_replace('"', '\"', $info_param) . "\"", $file_content, 1);
			elseif ($options == "sq")//separated & quoted
				$file_content = preg_replace("/$default/m", "$var \"" . str_replace('"','\"',$info_param) . "\"", $file_content, 1);
			elseif ($options == "sc")//separated & ending with a comma (used in JC2MP Example)
				$file_content = preg_replace("/$default/m", "$var $info_param,", $file_content, 1);
			elseif ($options == "sqc")//separated & quoted & ending with a comma
[...]

PR #44, opened 2017-02-21
Create config for Wurm Unlimited

smalls89:
Config xml for Wurm Unlimited


modules/config_games/server_configs/wurmu_win64.xml
```diff
@@ -0,0 +1,94 @@
...
+    <mod key="wurmu">
+      <name>none</name>
+      <installer_name>402370</installer_name>
+      <installer_login>anonymous</installer_login>
+    </mod>
+  </mods>
+  <server_params>
+    <param id="SERVERNAME" key="servername=" type="text">
+      <option>ns</option>
+      <desc>The name of the server that will be visible in server browsers. (NOTE: If the server name contains spaces, quotes are required! ie. "Server Name")</desc>
```
Zorrototo: you can use different options to do what you want, I think it is:

        nothing / no value = placed as is
        s = space / separated
        q = quoted
        sq = space and quotes
        sc = space and ends with a comma
        sqc = space, quoted, and ends with a comma

Zorrototo: Just my two cents :)

rocco27:
2 more things to make your xml nicer:
- Please do not set custom game name, follow the original game name. Example: Wurm Unlimited Server Windows is total wrong and unnecessery, use only Wurm, it is enought.
- You can set the hostname automaticly and we do this usually. The owner always set the hostname in the OGP, why we force them to do it twice?

own3mall:
Please make the changes recommended above, update your branch and pull request, and then I'll merge it.  Thanks!

modules/config_games/server_configs/wurmu_win64.xml
smalls89: True but i tried various options and i couldn't get it to work correctly, the cli param needs to have quotes if it has a space but also it cant have a separating space ie (servername="server Name")

this was the only way i could get it to work, i might have overlooked something though...

smalls89:
rocco
> Same with the admin password, this called RCOn generally and OGP can manage it
Could not find a correct query type? Why is it missing?

Im assuming these are refering to rcon in the way most games use it... Wurm does not have such a feature

the admin password is to be used ingame via the console

else there is the prompt window of the server, but you can't type in it...

rocco27:
Doesn't matter how can you manage RCON/Admin ingame, when I said OGP can manage it I thought you can set the password automaticly in the start up command line by OGP. If the game server support general RCON then it is other thing and maybe you can manage the RCON comands in the Log window from console.

rocco27:
Doesn't matter how can you manage RCON/Admin ingame, when I said OGP can manage it I thought you can set the password automaticly in the start up command line by OGP. If the game server support general RCON then it is other thing and maybe you can manage the RCON comands in the Log window from console

smalls89:
I image these ports can be set to anything, but these are defaults

Wurm is unique when i comes to admin, everything is done through an in game item oe in game console, there is NO external console prompt, rcon, or otherwise

Zorrototo:
Just a question, why do you have this post install? is it mandatory when you install a fresh server? The post install script is used after the Install or update of server.

smalls89:
I saw no alternative as the install puts down the 2 backup maps. If the user chooses to use the default maps, the server will throw the biggest fit about the maps not being found.


Side note with using q option, CLI separates it with a space, don't know if this is a bug/glitch, out I'm just doing it wrong...

Zorrototo:
You mean the "q" option for servername in server parameters is not working for you? be more specific to speed up the polishing of your xml :)

own3mall:
If you look at https://github.com/OpenGamePanel/OGP-Website/issues/19 there is only ns... adding new options now.

own3mall:
Hope I didn't break anything.  I documented the changes:

https://github.com/OpenGamePanel/OGP-Website/wiki/XML-Notes

If any of you want to help complete the XML documentation following a format similar to the one in place (unless you can come up with a better one), I'll grant you access.  We can add to it as confusion and clarification is needed.

smalls89:
so option q should work now???

Zorrototo:
as I said 

> edit4: also your VAC parameter seems to be not valid for Wurm Unlimited absolutely no reference in official doc, and even the log tells you it is wrong.

`Unknown parameter: -insecure`

own3mall:
Yes, "q" should work.

Zorrototo:
@own3mall if you can create a sub page in the wiki like "WIKI WIP" where I can import the current one and start complete it, test things and all (I'm new to GitHub and don't want to break anything on the real one :D) and grant me access to this sub-page, I'll start doing it. When I'll have it complete, polished and verified we can then import it in the real one and destroy this 'WIKI WIP" page.

smalls89:
I suppose this is done.... Unless there is a way to edit sqlite, all the remaining settings are in a sqlite db within the respective map folder...

Zorrototo:
Seems pretty good yes. Don't have the game to test but server starts good.

Can you change it back to win32 please, this way we keep the good organisation for ogp xml :D (some older xml don't  but for now better use the good naming convention)

Zorrototo:
Something you can also add is an icon for this game server, I think after that this can be accepted as a functional XML :) 👍

smalls89:
ih, i thought the win32 and win64 where for specific versions, wurm is indifferent, it works on either, and depends on the version of java installed locally, can change it back if it is still an issue...

smalls89:
idk how the icons work....

Zorrototo:
For your XML simply upload the icon to OGP Panel folder /images/icons/ name should match XML so wurmu.png will work.

own3mall:
Yay for making progress!  Let me know when this is ready.

smalls89:
believe thats about it...

PR #54, opened 2017-02-27
 Minor Fixes

Adjokip:
This is mostly fixes for warnings or notices that appear with max error
reporting. There's still othr stuff, but this'll do for now.

Form_table_class: Uses the old, PHP-4 constructor method which is now
deprecated.
Refreshed: Same as above, used the PHP-4 constructor method. PHP 5.3 and
above is supported. So, no need to use much, much, older method.

lib_remote: Set the default timeout in the constructor method to match
that of what was commited in 6e1629f , and changed the remote_dirlistfm
method from having a function within the public method to utilize the
callback arguement.

view: Added a better check for the installation file removal. $admins
will either be false or an empty array. This prevents the script
attempting to delete the file every page load if the file doesn't exist.

show_users: Added single quotes around the user_id key to prevent a
notice.

TS3Admin: getRemoteServers returns false is there's no servers. Added a
check to make sure there is remote servers - if there's not, display a
message. Added a translation.

FTP Users: getRemoteServers returns false is there's no servers. And as
such you cannot add FTP users. So display a message to the user
notifying them about this.

is_numeric being passed to array_walk in this fashion was always
returning true, however generates muliple warnings. Now, array_filter is
used and the count value of that is used. Which also removes the need
for the first check.

Status: Added a check for the GD extension for the bar.php images. If GD
isn't loaded, the images won't be displayed but the other components
are. A message will be displayed telling the user that GD is required.
Added translations.


lang/English/modules/settings.php
```diff
@@ -45,7 +45,7 @@
 define('feed_url', "Feed URL");
 define('feed_url_info', "GrayCube.com is sharing a LGSL feed on the URL:<br><b>http://www.greycube.co.uk/lgsl/feed/lgsl_files/lgsl_feed.php</b>");
 define('charset', "Character Encoding");
-define('charset_info', "UTF8, ISO, ASCII, etc... Left it blank to use ISO encoding.");
+define('charset_info', "UTF8, ISO, ASCII, etc... Leave it blank to use ISO encoding.");
```
Zorrototo: I'm not sure this string now reflects what it does. I think if it is empty, it uses what in defined in the lang file (which is now UTF-8 for all because of transifex).

Adjokip: It's used on the panel settings page when hovering over the Character Encoding tooltip.

'Leave it blank' seems more natural than 'left it blank' - but if the word left needs to be used, then I'd probably phrase it, 'If left blank ISO encoding will be used.'

Zorrototo: Yes, I was not talking about the better wording, but what I mean is that this string is surely not true now, because it uses what is defined in the main lang file when empty, which is UTF-8. It will not use ISO encoding when empty but UTF-8 which is the standard we now use with Transifex<->OGP.

Adjokip: It seems English/globals.php file uses ISO as lang_charset, a quick look at some of the other language globals, and they use UTF-8.

It'd probably be good to change that to UTF-8 also, just for the sake of consistency. Afterwards, the translation above could be updated.

Zorrototo: I don't even know if we still need this parameter actually. as we defined the encoding from lang files now (I think it was not like that long long time ago, will check sourceforge for that => yeah it was not like that on really old version). Maybe some inputs from own3mall on the topic will enlighten us more.

own3mall:
Must we declare our functions public?  I never decorate my classes that way.  

Encapsulation in the .NET world (the type of work I do professionally) is not very effective... in my opinion anyways.  

As long as everything works, I'll merge this.  I don't have the time to fully examine the changes since they appear very small,

own3mall:
But for the love of my sanity, please resolve the conflict?  I really don't want to do it.  I guess if I do, I will use TortoiseGit and Windows since I'm no expert on the git command line and had problems doing it last time from Linux...

own3mall:
I believe the encoding setting is still used.  As to whether or not it affects the language files anymore, I have no idea, but it should be putting an encoding entry in the header for the panel.

Zorrototo:
What I see in my Panel with French lang, and nothing in this parameter: `<meta http-equiv="Content-Type" content="text/html; charset=UTF-8">`
What I see in my Panel with French lang, and ISO in this parameter: `<meta http-equiv="Content-Type" content="text/html; charset=ISO">` and of course french accentuated words are messed up.

What I see in my Panel with English lang, and **nothing** in this parameter: `<meta http-equiv="Content-Type" content="text/html; charset=ISO">`
What I see in my Panel with English lang, and **ASCII** in this parameter: `<meta http-equiv="Content-Type" content="text/html; charset=ASCII">`
What I see in my Panel with English lang, and **LoL** (yeah I know 🤣) in this parameter: `<meta http-equiv="Content-Type" content="text/html; charset=LoL">`

So yeah it definitely works and change the page encoding in the header, but uses the encoding defined in the lang file when this parameter is empty.

Adjokip:
@own3mall Conflict should be resolved now.

Functions within a class default to public unless otherwise stated, however I think to be consistent with some of the other code, it's good to state that it is public - and it's generally just good practice to do so.

But everything does work - this commit was to fix some warnings/notices appearing at maximum error reporting.

PR #55, opened 2017-03-01
Added rust support in LGSL

acsid:
I fixed the mess ive done the last time.


Zorrototo:
That's correct but maybe sort it alphabetically in order to keep... order :D

rocco27:
What was wrong with GameQ? It works perfectly with Rust.
You can extend LGSL by Rust, but then do not forget to add lgsl_port_conversion($type, $c_port, $q_port, $s_port) as well!

Zorrototo:
I don't think there is something wrong with GameQ, but why not having both available? :) (personally I prefer LGSL :D)

acsid:
I added Rust to LGSL mainly to get it working into the LGSL server list. Let me take a look at that part.

Zorrototo:
It actually currently work with LGSL if you setup as a source server :) but yeah it is better to have it with proper name :D

rocco27:
Okey I got it.

PR #56, opened 2017-03-04
Hellion

smalls89:


smalls89:
Before merging, I would like some help.

In the replace_texts section there is an option for the query port, commented out however as it doesn't work

it needs to take the port and add 1 and i have no idea how to make this happen

rocco27:
If you remove those lines:
```
  <reserve_ports>
    <port type="add" id="QUERY_PORT">1</port>
  </reserve_ports>
```
and add between <server_exec_name> and <cli_template>
`  <query_port type="add">1</query_port>`
then you can define query by server port + 1 = query port and use it in the <replace_texts>.
```
  <text key="query_port">
    <default>status_port=.*</default>
    <var>status_port=</var>
    <filepath>GameServer.ini</filepath>
    <options></options>
  </text>
```

smalls89:
Couldn't get it to work...

```
[2] => Array
(
    [info_param] =>     <- Query port should go here but is blank...
    [default] => status_port=.*
    [var] => status_port=
    [filepath] => GameServer.ini
    [options] => 
)
```

Did a find all and only found this item (`<query_port type="add">1</query_port>`) in older games. Is it possible that somewhere along the line this item got hooped and doesn't work anymore?

rocco27:
<query_port type="add">1</query_port> should work, Rust xml use it.

smalls89:
I see... Rust has this line it, but query port is only used in the cli params, not replace texts.......

rocco27:
Yes, that is true, then check 7 Days to Die xml, we are using this and works fine.

smalls89:
i cant, i dont have a linux box to test it on...

rocco27:
What? You don't need Linux to check 7 Days to Die xml. Open the xml file and see the content. Anyway 7 Days to Die supports both Windows and Linux systems.

smalls89:
i see the default telnet port in the server config is 8081, the xml doesn't have a defining setting for query port, and the run script uses netcat to connect on port +1 to shut it down.....  i dont know how this one works either.... but i cant get this one to work...

rocco27:
The name does not matter again, OGP set the query and we use where we want. Query port from OGP used for telnet in 7 Days to Die, you can set it for query port in the ini file.

Zorrototo:
Maybe it needs modification to lgsl protocol file to have the query port variable working?

own3mall:
Is this ready yet?

smalls89:
Can't get the query port to work

rocco27:
I will try to install this gameserver tonight and see how query works and fix it for you. Right now I am busy with finishing and releasing ARK configs files.

rocco27:
1, Post install failed.
2, Query port should work! I see it is written correctly to the GameServer.ini file. You must able to connect to the gameserver and query it. OGP will query only if we set query in the xml file, now you havent set anything so how do you except any queres? We can extend GameQ or LGSL query protocols, but need to find a proper way.
3, Does it work under WINE? It doesn't start on my side.

I will investigate bit more later.

own3mall:
Great, hopefully we can merge and close this out soon.

PR #59, opened 2017-03-08
Assigning Multiple Cores to Servers

Adjokip:
Fulfilling a feature request posted [here](http://www.opengamepanel.org/forum/viewthread.php?thread_id=5285) by @FarukGamer - these small changes allow you to assign multiple cores to a server.

It requires a change in the ogp_game_mods table, changing the cpu_affinity column from a varchar(2) to varchar(64) (a maximum possible of 32 cores, as the core info is seperated by commas, eg, 0,1,2)

It doesn't require anything to be changed on the agent side - as the agent just uses `taskset`, which accepts cores the way I mentioned above.

It'll also attempt to verify that a valid core has been selected. As in, if you've got 4 cores, the user cannot edit the checkboxes and change the value of one to a non-existing core. Or edit the checkbox value to a
non-numeric value. Doing so will default the field to 'NA', which the agent handles.

It's mostly srcds servers I've hosted for in the past which don't benefit from multiple cores. Although, I could see this being useful for some games.

### DIFF
--- a/js/modules/user_games-mods.js
+++ b/js/modules/user_games-mods.js
@@ -6,9 +6,17 @@ $('.set_options').click(function(){
 	var mod_cfg_id = $(this).attr('id');
 	var modtr = $('#mod_cfg_id_'+mod_cfg_id);
 	var addpost = {};
+	
+	var cpus = [];
+
+	$("#cpu_select").find('.cpus:checked').each(function(i, e) {
+		cpus.push($(this).val());
+		addpost['cpus'] = cpus.join();
+	});
+
 	addpost[ 'set_options' ] = 1;
 	addpost[ 'cliopts' ] = modtr.find('#cliopts').val();
-	addpost[ 'cpus' ] = modtr.find('#cpus').val();
+
 	addpost[ 'nice' ] = modtr.find('#nice').val();
 	addpost[ 'mod_cfg_id' ] = mod_cfg_id;
 	if( modtr.find('#maxplayers').get(0) )
--- a/lang/English/modules/user_games.php
+++ b/lang/English/modules/user_games.php
@@ -212,4 +212,5 @@ define('add_folder', "Add folder");
 define('add_folder_info', "Write the name for the new folder, then click on the icon.");
 define('valid_user', "Please specify a valid user.");
 define('valid_group', "Please specify a valid group.");
+define('set_affinity', 'Set Server Affinity')
 ?>
--- a/modules/user_games/edit_home.php
+++ b/modules/user_games/edit_home.php
@@ -504,8 +504,35 @@ function exec_ogp_module()
 		{
 			$maxplayers = 0 + @$_POST['maxplayers'];
 			$cliopts = $_POST['cliopts'];
-			$cpus = $_POST['cpus'];
-			$nice = $_POST['nice'];
+
+			// Get the total CPU count. and, if the agent is offline, set the CPU to NA.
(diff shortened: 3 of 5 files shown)


FarukGamer:
I did wrote it already on the forum. Anyways, thank you for fulfilling this request 😃

own3mall:
Does this impact Windows?

Adjokip:
@own3mall Initially it impacted Windows. But it should work now.

Don't have a VM up for testing, but the output of `$result` is as you'd expect to pass to the /affinity flag.

FarukGamer:
Does this works with Linux?

Adjokip:
@FarukGamer Linux Mint is what I tested it on. So it should be working on those environments.

`while sleep 1; do ps -Uogp_agent -Lo pid,lwp,psr,comm | egrep 'PID' && echo -e "\n"; done` shows the cores being assigned properly.

FarukGamer:
Alright

own3mall:
Still need someone to test this on Windows, but I guess it's good for now.

FarukGamer:
Now I have updated the Panel and Agent without any problems. But strangely, the function is missing to assign it to multiple cores.

http://prntscr.com/eia9lc

Adjokip:
@FarukGamer Yeah. I accidentally removed this when working on the auto-update feature, after creating another branch. Previously I was just submitting changes to my master branch, which, evidently isn't very practical when working on multiple features :P

I resubmitted the PR in #61. Also got a chance to test it on Windows, works there.

FarukGamer:
Oh I have not seen it, sorry

FarukGamer:
It seems so that the selected boxes do not save properly. I have noticed it after I have reloaded the page.

Zorrototo:
I tried to assign different cores to a game server, and yeah they seem to not save when your refresh the edit page, none is selected. Database however has a value saved so I guess it works.

Note also that there is currently a bug with the module installation/update so be sure to check that thread http://www.opengamepanel.org/forum/viewthread.php?thread_id=5295 or your database will surely not have the proper modification to store the changes to cpu affinity. I also see the taskset command in the server startup script so I think it works properly.

The problem is that it does not tick the already set CPU so this should be changed if possible to reflect the current selected CPU.

FarukGamer:
taskset: failed to parse CPU list: 0, taskset: failed to parse CPU list: 0, taskset: failed to parse CPU list: 0, taskset: failed to parse CPU list: 0,

FarukGamer:
taskset: failed to parse CPU list: 0

This is what I get now

Adjokip:
@FarukGamer That'll be because the query failed to run from modules.php.

Modify it manually in the meantime:
```sql
ALTER TABLE `ogp_game_mods` modify column `cpu_affinity` varchar(64) null AFTER `extra_params`, comment = 'utf8mb4_general_ci';
```

It'll be fixed once @Zorrototo's PR is accepted. The query fails because of a syntax error. Which I should have noticed. Sorry.

Adjokip:
That's because the query fails to run from modules.php.

It'll work once @Zorrototo's PR is accepted.

In the meantime, you can run this to fix it:

```sql
ALTER TABLE `ogp_game_mods` modify column `cpu_affinity` varchar(64) null AFTER `extra_params`, comment = 'utf8mb4_general_ci';
```

I should have noticed the SQL error in modules.php. Sorry. If you notice the value is 0, ... so it's still at VARCHAR(2) rather than VARCHAR(64).

PR #60, opened 2017-03-09
Auto update

Adjokip:
Fulfilling the request [here](http://www.opengamepanel.org/forum/viewthread.php?thread_id=5250) - this adds the ability to do auto updates, get the latest buildid, and get a game's current build id.

Tested with TF2, Ark, and GMOD.

### DIFF
--- a/includes/lib_remote.php
+++ b/includes/lib_remote.php
@@ -167,6 +167,8 @@ class OGPRemoteLibrary
 		if ( $response === NULL )
 			return 0;
 
+		if ( $response == -10 )
+			return 'Agent Returned: -10. Home not found.';
 		@list($retval,$data_tmp) = @explode(";",$response);
 
 		// We get log only with positive values.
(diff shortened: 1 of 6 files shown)


own3mall:
I believe we can expose the auto-update functionality via the panel with a script that accepts the game server's home ID and accepts a uniquekey passed in via GET parameters.  I'd be happy to implement this part.  Basically, I'm thinking a button near the manually initiated SteamCMD update functionality that shows a link the user can copy to access the auto-update functionality for their server which can then be used in a cronjob using wget.

Thoughts on this?   This code will not require a login... just a key that only the admin or the user knows.

own3mall:
And I'll add an action to the cron module to support auto update as well.

Adjokip:
I think that'd be a good idea. I already posted a mockup script that accepts a home_id on the forums, [here](http://www.opengamepanel.org/forum/viewthread.php?thread_id=5250#post_25661) - it could also be edited to have a unique key passed to it - and it already has the `php_sapi_name()` function call, so only allows CLI access.

That was just a quick example though - I'm sure something better could be done from what I posted.

PR #61, opened 2017-03-10
Multiple Core Assignement

Adjokip:
Accidentally removed these when creating a branch for the auto-update
feature. Woops.

But I was able to test it now on Windows. Previously only on Linux Mint.
Can confirm it works on both... only needed a slight change for Windows.


FarukGamer:
Oh I have not seen it, sorry

FarukGamer:
@own3mall Is this not gonna be merged?

PR #65, opened 2017-03-11
Remember Checked Affinity Boxes.

Adjokip:
As per @FarukGamer and @Zorrototo's suggestion, this pull request will put the 'checked' attribute on each checkbox which are currently assigned to a server.

I added a new database method so the data may be used elsewhere if so desired.

Tested on Windows and Linux Mint.

### DIFF
--- a/includes/database.php
+++ b/includes/database.php
@@ -66,8 +66,8 @@ abstract class OGPDatabase {
     abstract public function getGroupList();
 
     abstract public function getUsersGroups($user_id);
-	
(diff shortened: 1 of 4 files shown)


Zorrototo:
Applying this breaks the Panel. Blank page with no error. database.php is the cause apparently

Adjokip:
@Zorrototo Maybe you forgot to change database.php or it got corrupt? I can't get any errors after applying every change manually. 

Ok, I'll include those changes now.

PR #78, opened 2017-03-17
making pagination for user list, show homes, server monitor, and fix logger ordering

Obada8:


own3mall:
Looks OK.  I will add a user preference to the account settings page to set the limit soon.  That way, if a user doesn't want to be limited, you could set a value like 9999999999999 or something like that.

Thanks for the work!

Adjokip:
@own3mall, @Obada8, This commit is missing content?

There's nothing under _Files changed_, the commit was reverted shortly afterwards - and as such, there's no changes in the master branch.

Obada8:
i making mistake with these when i try to add pull request for open game panel on GitHub :\ ( i am still noob here )

own3mall:
How interesting.  Github initially showed the commit order in reverse...

PR #86, opened 2017-03-18
Adding spaces in pagination to avoid the one line..

Zorrototo:
..breaking page if you have more than 45 pages to show.


Adjokip:
`rtrim`, a few lines below, will also need to be changed.

`echo rtrim($pagination, ",");` to `echo rtrim($pagination, ", ");`

Otherwise the last comma will remain.

Zorrototo:
You're right :) thanks.

PR #89, opened 2017-03-22
Fix BUG in last pagination and add Pagination For Dashboard

Obada8:
in last pagination i was missing prefixes between mysql & mysqli and add
new pagination for dashboard


own3mall:
Please add checks for the user preference when limiting items per page.

Obada8:
i will do it when i go to my home, thanks for inform

Adjokip:
@own3mall @Obada8 This won't work as it is and the following should be fixed before accepting.

`getHomesFor_limit` takes 4 arguments, but you're only passing 2 to it. (home.php, lines 148-150)
`getIpPorts_count` takes 2 arguments, but on dashboard.php line 103 you're only passing 1 argument to it.
`getIpPorts_limit` takes 3 arguments, but you're not passing anything to it. (dashboard.php, line 53)
`getIpPortsForUser_limit` takes 3 arguments, but you're only passing 1 to it. (dashboard.php, line 59)

If you're not passing the set amount of arguments to your function, warnings are going to happen. Either declare a default in the function definition or pass a value to the function.

Additionally, I'd suggest checking to make sure `$_GET['page']` is an integer before passing it to your function, as I did [here](https://github.com/Adjokip/OGP-Website/blob/f261b098ce0c83fb8cc69d6cb3ce8e020c1928e3/modules/user_admin/show_users.php#L50).

Otherwise, what happens if a user changes the value of page to a non-numeric or value less than 1? You'll get an SQL error.

This won't work because some of those missing arguments are used in queries, but they don't exist. So the limit statement probably becomes `LIMIT 0, NULL`. resulting in an SQL error.

Obada8:
ok, i will close this request too, and upload new request with same title, thank you guys :)

PR #90, opened 2017-03-22
list servers in bar menu according to current page & limit

Obada8:
list servers in bar menu according to current page & limit


Zorrototo:
This is already in your previous PR https://github.com/OpenGamePanel/OGP-Website/pull/89/files

Obada8:
thanks for inform, i will close it

PR #91, opened 2017-03-23
Fix BUG in last pagination and add Pagination For Dashboard

Obada8:
Fix BUG in last pagination and add Pagination For Dashboard

### DIFF
--- a/home.php
+++ b/home.php
@@ -138,7 +138,7 @@ function ogpHome()
 {
     global $db,$view,$settings, $loggedInUserInfo;
 
- 	$page_user = isset($_GET['page']) ? $_GET['page'] : 1;
+	$page_user = (isset($_GET['page']) && (int)$_GET['page'] > 0) ? (int)$_GET['page'] : 1; // thanks for Adjokip
   	$limit_user = isset($_GET['limit']) ? $_GET['limit'] : 10;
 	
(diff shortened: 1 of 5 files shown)


Adjokip:
dashboard.php line 108: `getIpPorts_count` is still missing the second argument, which will result in a warning when show errors is on because `getIpPorts_count` doesn't have a default value (even of '') for the second arg.

It'd be good to provide it just for the sake of consistency... or change the function declaration so the second argument isn't required; thus a warning won't be shown.

And I'd still suggest adding a check to `$_GET['page']`, as can be seen [here](https://github.com/Adjokip/OGP-Website/blob/f261b098ce0c83fb8cc69d6cb3ce8e020c1928e3/modules/user_admin/show_users.php#L50) - because you're always expecting it to be an integer, anything else such as 0 or a string should default to 1.

Obada8:
yes you are right. i will fix it :|
I do not know what happened to me,

thank you adjokip

Obada8:
fixed.

own3mall:
My comment disappeared?  Make sure you check for user's account preference for the number of items shown on each page.

own3mall:
Never mind, I see you opened a new pull request.  Looks OK to me.

Obada8:
yes, i was closed last request and i open new request with same title.

PR #123, opened 2017-04-30
Update

SanKen:
Fixed tables. The white background is gone.


Zorrototo:
it breaks my ftp admin page.

Zorrototo:
should be closed as no reply from author and a simple check by yourself will confirm that it breaks Panel.

PR #124, opened 2017-05-09
Fix Core Assignment For Games Under Wine

Adjokip:
The what_os() call includes 'wine' in the returned string, thus is was being caught by the preg_match and what was being stored was the Windows bitmask for the affinity.

Checking for Cygwin seems to be correct from looking over what's returned by what_os().

Fix for: http://www.opengamepanel.org/forum/viewthread.php?thread_id=5389 and working on Ubuntu/Windows 2012


own3mall:
That is also what I'm seeing as the return value from uname in cygwin.

PR #125, opened 2017-05-10
Rising Storm 2: Vietnam support

Zorrototo:
Early support for the coming beta wave.

For now it does the basics. Will see how to improve it in the future when documentation will be released.


Zorrototo:
Will not mess with LGSL files for now as it perfectly works with the Red Orchestra 2 protocol, but will need to do it for better support from other modules like LGSL_with_img_mod (to list it as 'Rising Storm 2: Vietnam' instead of 'Red Orchestra 2' :D)

PR #129, opened 2017-05-16
Updates to ARK: Survival Evolved and DayZ Mod / Arma 2

ProGEEK:
Updated ARK: Survival Evolved to use correct query ports
Fix LGSL for DayZ server templates
Remove Underscore from lgsl_query_name since this gets stripped out anyway in lgsl_software_link()
Added basic DayZMod Operation Arrowhead server template


rocco27:
For me DayZ means DayZ Standalone. If it is mod, then better to keep dayz_arma2co lgsl name instead of dayz only.
What we do if once they release DayZ Standalone server files? DayZ Arma2 mod already use dayz name, so it just makes mess only...

rocco27:
Why did you changed ARK lgsl? Now it will make big mess for everyone.
Now you just changed the query from 7778 to 27015. Query doesn't matter, can be use both. But now you also set $c_to_q = 0 which means server port 27015 and query port 27015 which is wrong. These two should be 2 different and players connect to query port. In ARK and many "new" gameservers started to use query port to connect istead of server(connection) port somewhy...

ProGEEK:
Please read http://ark.gamepedia.com/Dedicated_Server_Setup#Network
ARK by default requires 3 separate ports, Game port, RAW UDP port and Query port.
7777 is the default Game client port (This doesnt seem to actually get used by ARK), 7778 is not the query port, it is in fact the Raw UDP socket port (always Game client port +1), 27015 is the Query port for Steam's server browser
This is what my server reports when told to start with the default port (7777) and the correct query port
https://pastebin.com/hkTp5bmh

Setting $c_to_q to anything other than though takes server port and adds one, in the case for ark, line 466 is triggered:  if              (!$q_port &&  $c_port && $c_to_q)       { $q_port = $c_port + $c_to_q; }
Which overrides the default query port and forces it to 7778 (Incorrect)
Setting $c_to_q to zero caauses line 528 to be triggered: elseif  (!$q_port &&  $c_port && $q_def)        { $q_port = $q_def; }
Which actually sets the query port to the default port (In ARK's case, 27015)

ProGEEK:
dayz is actually incorrect, it shouldve been dayzmod (I have another commit coming)
It was changed to conform with other services that Open Game Panel references externally (For images etc) such as Gametracker: https://www.gametracker.com/games/dayzmod/ for the LGSL "Extra" (Map Icons etc)

rocco27:
Okey, let's change the default query port to 27015, but then use $c_to_q = 19238, so If you set  7778 to game port then query will be 27016, calculated automaticly.

ProGEEK:
Works for me

rocco27:
Hmm, I already rechecked everything and something is wrong.
You have right. In the official documentum the game port is 7777, the game raw port is 7778 which is not used and the query port is 27015. That's clear, but we can use different ports anytime for anyone and I think the reason is why I changed to use 7777/7778 was the calculation for LGSL.
In the XML we have to define how to calculate the query port from server/game port. It was server port +1, I set same in the LGSL. Then query worked perfectly even if you had multiple ARK game servers. It was also easy to set the server port to anything. If you set 30000, then query was 30001.

If we want to have 27015 as default query port, then we have to use server/game port +19238 in XML and LGSL too. For only one server is good, but for multiple ARK servers makes problems.
Example want to use 27015 for default query in LGSL and XML. If we use multiple ARK server then
1, Game port 7777, query port 27015 (7778 port is reserved)
2, Game port 7779, query port 27017 (7780 port is reserved)
3, Game port 7782, query port 27019 (7783 port is reserved)
4, Game port 30000. query port 49238 (30001 port is reserved)

Is it fit for you, for everyone?

ProGEEK:
Either that, or we just have it add on 2 ports for query instead of 1, Im not sure if this will work from the ark browser, I will check tomorrow and confirm

rocco27:
Sure, I understand you and thanks for talking about that.
Unfortunatelly the "new" game servers don't follow old standard things. Like now with the ports. We have game, query, rcon ports seperatly. Before one gameserver used one port only. Hard to make a rule to our system to manage it automaticly.
On some game servers you have to connect to query port instead of server/game port.
Etc.

ProGEEK:
I have updated the LGSL protocol and the game xml files

own3mall:
Is this complete and ready to be merged?

ProGEEK:
If you are happy with the changes ive made, then yes

PR #132, opened 2017-05-21
Added support for Dont Starve Together dedicated server (Linux)

ProGEEK:
Added Basic XML and icon for Dont Starve Together dedicated server (Linux)


Zorrototo:
This game is hard to properly support, or support can only be limited like with your XML, no possibility of having multi cluster servers (I made XML to do that available on OGP forum http://www.opengamepanel.org/forum/viewthread.php?thread_id=4119 but it was kind of a mess in the end).

ProGEEK:
Yeah I got that feel when I was writing mine. But still, limited sort is much better than no support right ?

I'd be keen to add any changes/enhancements your XML file has that mine lacks though?

Zorrototo:
Thing is I never wanted to upload my set of XML files (there are two, one for Master one for Salve/Caves) because it was more a trick than a proper way to implement it in OGP. I was waiting for better system from devs, and didn't eared any change of this system yet, but I did not really follow development these last months.

ProGEEK:
My intention with this was so that anyone could get a simple server up and running. Clustering and caves (Which are beyond even me) will never be covered by my scope of XML

Zorrototo:
I'll give your XML a try tomorrow to confirm everything is working.

ProGEEK:
I have borrowed your custom params and post_install section (And tweaked a little).
Just a note that mine is intended as a standalone only. I.e. no sharding

Zorrototo:
You should not insert your personal cluster token in the post install script, this is a private key based on your personal game account. This is like sharing your  Steam CS GO token. :\

>  Server Tokens
In order to run a public Dedicated Server, GSAs are required to enter their “cluster token”. This token proves ownership of Don’t Starve Together and allows Klei to moderate servers that violate their terms of service. 

Also you should use an absolute path for persistent_storage_root see here http://forums.kleientertainment.com/klei-bug-tracker/dont-starve-together/linux-dedicated-server-bug-with-persistant-storage-folder-r309/?do=findComment&comment=345

Zorrototo:
Also you NEED to have the config file present right after installation (why I generated them in my XML) or else the custom fields will be applied on a empty file, then server will use the config file generated from custom fields and so things will be messed up.

ProGEEK:
There is no cluster token in the post_install script. The echo ... base64 line generates a basic cluster.ini file as per below
`
[GAMEPLAY]
game_mode = endless
max_players = 64
pvp = false
pause_when_empty = false

[network]
cluster_name = Dont Starve Together
cluster_description = A Dont Starve Together Server
cluster_intention = cooperative
cluster_password =
tick_rate = 30
autosaver_enabled = true
connection_timeout = 8000
server_save_slot = 1
enable_vote_kick = false

[account]
dedicated_lan_server = false


[STEAM]
DISABLECLOUD = true


[MISC]
console_enabled = true
autocompiler_enabled = true
`

I found with using {OGP_HOME_DIR}, when the server started, it would actually create a folder called `{OGP_HOME_DIR}` (Not sure why, seems like it was not actually replacing it, yet later in the process, it actually would, so I dont know

Zorrototo:
ho my bad I overlooked :D and yeah, I looked again and I used ../ too in my XML so this is actualy working too.

ProGEEK:
Now that I actually say that about the OGP_HOME_DIR and change it back, its stopped doing it.. odd

Zorrototo:
base64 is cool but **if** this is not something found on every distro OGP supports you shouldn't use it.

ProGEEK:
base64 should be pretty core to most systems, There will be cases where it is not installed (Custom installed system, minimal etc), but should be available via package manager

Zorrototo:
OK then :)

ProGEEK:
I changed the persistant_storeage_root back to using {OGP_HOME_DIR}.

The quirky behaviour I was seeing on my ssytem where it was actually creating a folder called `{OGP_HOME_DIR}` seems to have ceased

own3mall:
That's a reserved variable that gets converted in the agent.  It should always work.

Zorrototo:
Didn't have time to test it, hopefully everything works :D

PR #134, opened 2017-05-25
Removed -netlog in CLI

Zorrototo:
Because it crashes servers it is no longer working CLI parameter


Zorrototo:
thread about the issue here http://www.opengamepanel.org/forum/viewthread.php?thread_id=4894 or here http://www.opengamepanel.org/forum/viewthread.php?thread_id=5378&pid=26889#post_26889 or here http://www.opengamepanel.org/forum/viewthread.php?thread_id=4894

PR #136, opened 2017-05-27
helper mod for website

SanKen:
http://www.opengamepanel.org/forum/viewthread.php?thread_id=5410


own3mall:
I went ahead and added this functionality because it is helpful to retrieve the IDs used in the panel to match up some of the files created by the agent.  Closing this request.

PR #140, opened 2017-06-02
Make it consistent to also use TOP for Webhost Status...

Zorrototo:


Zorrototo:
http://www.opengamepanel.org/forum/post.php?action=edit&forum_id=2&thread_id=5433&post_id=27103

PR #141, opened 2017-06-02
English language and TeamSpeak 3 PHP Framework updated

rocco27:
The changelog for TeamSpeak 3 PHP Framework is not full in the commit, GitHub already cut it. Here is again: [https://github.com/planetteamspeak/ts3phpframework/blob/master/CHANGELOG](https://github.com/planetteamspeak/ts3phpframework/blob/master/CHANGELOG)


rocco27:
One more note, the version numbers in the files are wrong. It shows 1.1.24 instead of 1.1.25. I copied original files and had no any edit. I follow the dev, so will send another update once new version released.

PR #146, opened 2017-06-05
I just forget to add those line on previous pull request.

Obada8:


Zorrototo:
What is this for? even if people can understand it I think that would be a good thing the had title or description better than this.

Obada8:
@Zorrototo I'm sorry for making two title, this happened because i made this pull by GitHub site not from software, this pull request is just  a complement for previous pull request [139](https://github.com/OpenGamePanel/OGP-Website/pull/139).
I'm surprised why this still pending not merged yet :\ , i take to @own3mall to make this pull before add this here.

own3mall:
It could use a better description.  I just haven't been online.

PR #149, opened 2017-06-07
Missing a string from the resource

rocco27:
Link: https://opengamepanel.org/forum/viewthread.php?thread_id=5057&rowstart=80#post_27186


rocco27:
Wait, something is wrong with this. I can not see the text in the panel.

rocco27:
Everything was okey, but I found an other sting which was lost too. Ready to pull.

PR #150, opened 2017-06-07
Removed Bootstrap and added Jquery Datetime Picker

oNdsen:


own3mall:
This assumes we want to get rid of Bootstrap.  I don't care either way though, so it looks fine to me.

PR #153, opened 2017-06-12
Adding Display Public IP override Option for Gameservers behind NAT

oNdsen:


own3mall:
What's the point of this pull request, and why is this needed?

oNdsen:
[http://www.opengamepanel.org/forum/viewthread.php?thread_id=5453](url)

PR #154, opened 2017-06-12
Adding search function.

Obada8:
Hello every one, i made small query join to pagination count and select
value of certain search for watch_logger.php, show_homes.php,
show_users.php.
But we need to add search function to server_monitor.php too, but i
still faced some issue with display search result (Count for search
result not have problem with counting) by using this (in admin power),
search for function in pull request on `database_mysql.php `&
`database_mysqli.php`
` public function
getHomesFor_count($id_type,$assign_id,$home_cfg_id,$search_field){
if ( $id_type == "admin" )
{ `

and

`public function
getHomesFor_limit($id_type,$assign_id,$home_page,$home_limit,$home_cfg_id,$search_field){`

For this line:
`WHERE %1$suser_homes.user_id = %2$d '.($home_cfg_id ? 'AND %1$sserver_homes.home_cfg_id = '.$home_cfg_id.''.($search_field ?' AND %1$sserver_homes.home_cfg_id = '.$search_field.'' : '') : ''.($search_field ?' AND %1$sserver_homes.home_cfg_id = '.$search_field.'' : '').'').' ORDER BY home_id ASC LIMIT '.$gethome_page_forlimit.','.$home_limit.';';`

that appear on `else if ( $id_type == "user" ) `& `else if ( $id_type == "group" )` & `else if ( $id_type == "user_and_group" )`

this lines just a test from me if the statement work or not so will be ignore it at these moment.

So let's do this together then accept this pull request.

### DIFF
--- a/includes/database.php
+++ b/includes/database.php
@@ -178,7 +178,7 @@ abstract class OGPDatabase {
 
     abstract public function getHomesFor($id_type,$assign_id);
     
-    abstract public function getHomesFor_limit($id_type,$assign_id,$home_page,$home_limit,$home_cfg_id,$search_field);
+    abstract public function getHomesFor_limit($id_type,$assign_id,$home_page,$home_limit,$home_cfg_id);
 
     abstract public function getHomeMods($home_id);
 
--- a/includes/database.php
+++ b/includes/database.php
@@ -63,7 +63,7 @@ abstract class OGPDatabase {
 
     abstract public function getUserList();
     
-    abstract public function getUserList_limit($page_user,$limit_user);
+    abstract public function getUserList_limit($page_user,$limit_user,$search_field);
 
     abstract public function getGroupList();
 
@@ -178,7 +178,7 @@ abstract class OGPDatabase {
 
     abstract public function getHomesFor($id_type,$assign_id);
     
-    abstract public function getHomesFor_limit($id_type,$assign_id,$home_page,$home_limit,$home_cfg_id);
+    abstract public function getHomesFor_limit($id_type,$assign_id,$home_page,$home_limit,$home_cfg_id,$search_field);
 
     abstract public function getHomeMods($home_id);
 
@@ -256,7 +256,7 @@ abstract class OGPDatabase {
 
     abstract public function getGameHomes();
     
-    abstract public function getGameHomes_limit($page_gameHomes,$limit_gameHomes);
+    abstract public function getGameHomes_limit($page_gameHomes,$limit_gameHomes,$search_field);
 
     /// \return true If username and password match.
     /// \return false If username and password does not match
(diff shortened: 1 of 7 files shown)


Obada8:
@own3mall  You can accept this pull request now, when i finished from search field for game monitor i'll open new pull request.

own3mall:
Put the form submit back to 
```
onsubmit="event.preventDefault();"
```
It was probably done for a reason.

I'll try this pull request, but it seems like the queries have become very complex now with your changes.  I don't like complex SQL queries that become huge like that.

own3mall:
Also, why are you still having to use a distinct?  Just don't join the table that causes the duplicate entries due to different users?

Obada8:
@own3mall If i don't make join to users home then i can't use search by username, just who owner this server, btw, i'm not finished yet. I'm work on users group and sub users, when i finished i'll explain everything. i think you are confused because if statement `'.($search_field ? 'TRUE': 'FALSE' ).'` and i add new code in true and false area.

For `onsubmit="event.preventDefault();"` I'll try to avoiding delete this.

Thanks.

Obada8:
Database changes is done now, but for `onsubmit="event.preventDefault();"` if i keep it 
the `action="home.php"` will not work, any suggestion?

own3mall:
Why not just separate your queries?  Make one query to get a list of usernames that match up to the userid... then if you have a match, you can cut out the join in your new search query.

own3mall:
And why don't you add your new param to the post params on the form?  Then event.preventDefault(); should work just fine?

Adjokip:
The ternary operator seems a bit overused here... and maybe that's why own3mall thinks it's looking a bit complex - it does become a bit unreadable to keep track of things.

I would suggest not returning `resultQuery()` straight away, but rather building up the query string in a variable, and then passing that variable to `resultQuery()` at the end.

A quick example, no idea if it works: https://gist.github.com/Adjokip/406258dd7dc6301e426e38dc9af45c55

I wouldn't use the ternary operator in the way you're doing so here, but for smaller things conditional checks it's fine.

Obada8:
@own3mall Can i add just comments? OR do adjokip method? & What you mean by 
> And why don't you add your new param to the post params on the form?

& For 
> Then event.preventDefault(); should work just fine?

Yes it's work with no problem.
(The search function is finished for me & Ready for accepted.)

own3mall:
Yeah, I'm not a fan of the existing tables in OGP already.  The relationships are difficult to understand especially when the UNION command is used in so many of the queries.  The database wasn't designed as well as it should have been.  Oh well.

own3mall:
I just want to clean these queries up because as I said, I struggle with most of the existing ones already.

Obada8:
So what am I going to do now?

own3mall:
I prefer adjo's approach.  However, if you really don't want to do that, I will probably just merge it.

Obada8:
I'll do it, i just feel a little laziness :p

PR #159, opened 2017-06-15
Additional Check Display Public IP Support

oNdsen:
Also added a small colspan Fix on mini Start when an Error exists

### DIFF
--- a/modules/dashboard/dashboard.php
+++ b/modules/dashboard/dashboard.php
@@ -144,11 +144,10 @@ function exec_ogp_module()
 					}else{
 						$ip = $server_home['ip'];
 					}
-					if(ip2long($server_home['display_public_ip'])){
-						$ip = $server_home['display_public_ip'];
-					}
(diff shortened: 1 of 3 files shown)


own3mall:
Why the colspan change?   Screenshot of why you changed it?

oNdsen:
You only see it when you have Table Styles:
[https://picload.org/image/riwglopw/without.png](url)
[https://picload.org/image/riwglopw/without.png](url)

own3mall:
Um both URLs are the same.

oNdsen:
Oh damn... ;-)
[https://picload.org/image/riwglowr/with.png](url)

oNdsen:
Other Option is to add an empty <td/> but i prefer the Colspan.

own3mall:
Hope this doesn't affect the default theme negatively...

oNdsen:
Nope it doesnt. Allready tested it.

PR #166, opened 2017-06-21
Changing String and adding Display Public IP Info

oNdsen:


Zorrototo:
I guess you can add the lang file correction too, to delete the not used string.

rocco27:
String deleted.

PR #171, opened 2017-07-02
TS3 Module Fixes

Adjokip:
Updates the Smarty template engine, and also set the template path in ts3webinterface.class.php - as Smarty wasn't able to find the directories previously.

Also added a port field in the database instead of trying to connect to the default 10010 port so servers running on non-standard ports shouldn't cause errors when attempting to use the module.


own3mall:
Looks great.  Thanks!

tamitik:
Does this actually work. I just needed to change back to query rather than using the +24 variable that it was changed too. (jconnors)
 Reference the following in forums: http://www.opengamepanel.org/forum/viewthread.php?thread_id=5536&pid=28025#post_28025

PR #175, opened 2017-07-03
These strings are not used anymore I think

Zorrototo:


Adjokip:
These seem to be used on the server monitor page when you toggle 'Show Server IDs on Game Monitor page' to 'on' at the panel settings: https://github.com/OpenGamePanel/OGP-Website/blob/10952719af4dcf16335ddbaef602efc8114b7fba/modules/gamemanager/server_monitor.php#L225

Zorrototo:
Oh yeah my bad :D

PR #177, opened 2017-07-05
Update module.php

Adjokip:
The port column wasn't being added to the database due to the version not being updated in the last commit.

Maybe this will also fix @rocco27's issue? If he can check what his table structure is already like.


rocco27:
Now it is possible to assign back the ts3 server in game monitor.

Issue #180, opened 2017-07-06
New search fields problems


Zorrototo:
Two connected issues: 

When you are in Admin User list, and you search for a partial name (let say I have a user "Vodka" and I search for "Vodk"), it does not return any result. I tried also the 'joker' character but no luck. Partial searches are not supported.

When you are in Admin Game Servers list, I simply can't find a way to have a result (OK with empty search field it returns the whole game servers list, or if you put the IP too it returns the whole list if you have one IP only, or if you put the server ID it also works), whatever I try (part of server name, .....) I can't have any result, it always returns 'No game servers found' (by the way this message is not displayed as a 'failure' thing in red but as standard text).

Problem here is that partial searches simply do not work you need to have exactly the right thing written or it fails.

It is the same for the other search fields elsewhere. partial searches do not work

While testing I found that bug too: in game monitor, if you search for example one word from the server title, the page returns this error, which is obviously unrelated to the search: `No game servers assigned for you. You need to ask your OGP admin to assign games for you.` I am an Admin on my Panel and have all servers visible at Game Monitor.

Obada8:
I need to add another issue here, when i search for server with another owner username the search not showing the results. For example: my username is Administrator [Admin role] and i owned all game server else home_id = 84 [Owner name is test] when i search for this home_id via type ip for this home_id the search result show all game server which have this ip, but not show this home which had same ip.

own3mall:
@Obada8 Please be more specific.  How do I reproduce your page number problem step by step?  I don't understand what you're saying.

Obada8:
@own3mall I'm very sorry for my bad english :\ .
I mean try to make game server like [screenshot](https://s17.postimg.org/wbngr03cv/chrome_2017-07-08_03-39-15.png).
And go to list all game servers and search via ip only.

own3mall:
I believe all you would have to do is update the paging links to include the search field input as part of the URL if there is any?  Then it would work?

Obada8:
Yes, but this is wrong because displaying the page number in the link should return to the search results and not to the display of the game server (before click search).

own3mall:
@Obada8 I still don't understand what your problem is.  Is it still broken in the latest version of the OGP panel?  Update your panel to test.

Obada8:
@own3mall  I use latest update yes (just leave GitHub field empty in panel setting then click update)
anyway does search function work with you? even if game server not showing in first page?

Adjokip:
[...]

PR #181, opened 2017-07-06
Search Field and other minor fixes

Adjokip:
Fixes the issues noted by Omano in #180.

Also now checks if php-zip is installed on the update module. Previously, if you attempted to update without this extension you'd receive a warning.

Additionally fixed not being able to set an expiry date on user accounts as mktime() was returning false on PHP >=7 due to the $is_dst arg being removed - and had been deprecated since 5.1.0.

### DIFF
--- a/modules/user_admin/show_users.php
+++ b/modules/user_admin/show_users.php
@@ -100,7 +100,7 @@ function exec_ogp_module() {
         $user_expires = read_expire($row['user_expires']);
         print "<tr class='tr".($i++%2)." ";
         print $row['users_role'] . " ";
-        if(!empty($ownedBy)){
+        if(!empty($ownedBy) && empty($search_field)){
 			print "hide";
 		}else{
 			print "subusersShowHide";
--- a/modules/user_admin/show_users.php
+++ b/modules/user_admin/show_users.php
@@ -46,7 +46,7 @@ td.actions{
  */
 
 function exec_ogp_module() {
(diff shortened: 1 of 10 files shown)


Obada8:
Oh, i was tried many time to replace `=` by `LIKE` i failed in displaying result and i got error but i success in counting pages with `LIKE` i don't know why that error was  happened to me i see in you're edit you add multiple `LIKE `, So that reason i was keep the search on `=`.
Thank you.

modules/user_admin/show_users.php
```diff
@@ -55,22 +55,40 @@ function exec_ogp_module() {
...
+					<form action="home.php" method="GET" style="float:right;">
+					<input type ="hidden" name="m" value="user_admin" />
+					<input name="search" type="text" id="search" />
+					<input type="submit" value="'.get_lang('search').'" />
+					</form>
+				</td>
+			</tr>
+		</table>';
+
     echo '<table class="userListTable center" style="width: 100%;margin-top:100px;">';
```
Zorrototo: would change the margin top line 85 to 50px now everything is not on top of each other, and that would match the margin bottom for the show_homes page.

Adjokip:
Not a bug.

Those are passed to sprintf(), which requires you to escape the % character. You escape the % character in sprintf by adding an extra % character.

http://php.net/manual/en/function.sprintf.php

PR #183, opened 2017-07-08
Fixes for Searching

own3mall:


own3mall:
Could possibly address some issues discussed here:  https://github.com/OpenGamePanel/OGP-Website/issues/180

@Adjokip @Zorrototo can you please review / test this changeset?  It was working in my environment, but maybe I missed something... 

I am not a fan of the new search functionality.  We allow multiple database fields to be searched which is not how it should work.  Instead, the user should specify which column to search... but whatever.  I hope this works for now.

Adjokip:
Tested it - all working.

I agree the search feature should be improved so it would be possible to specify the search column - it would also let the user know what they can search.

own3mall:
Agreed, maybe @Obada8 would be willing to work on it since he introduced the feature :smile:

PR #195, opened 2017-08-02
Fix for languages where text takes two lines in the button

Zorrototo:


own3mall:
What happens if the height is removed altogether?  The button should auto-adjust in size?  We could set a max-height of 100px, but first try it without the height rule.

Zorrototo:
It is not good by removing the height property because then you have buttons with different heights.

own3mall:
And that's a problem because? HTML is usually pretty responsive when no height is set.  Button text should be kept relatively short.  Now because of other languages, we have to increase the height of all buttons to accommodate?  I'm fine with merging this change, but I'm not a fan of fixed pixel heights.  :D

Zorrototo:
It is a problem because it is visually disgusting.

Zorrototo:
By the way, same problem with english lang actually, for example the string "Update from local Master Server" takes two lines too.

PR #199, opened 2017-08-16
Adding latest commits description when update is available

Zorrototo:
thanks to Adjo for the code. Will need localization later if accepted (header text, and the word 'committed').


Zorrototo:
https://www.dropbox.com/s/3378hcjiltng19j/Capture%20d%27%C3%A9cran%202017-08-16%2019.28.41.png?dl=0

Zorrototo:
Added a check for Curl and move  the break to proper place.

PR #201, opened 2017-08-17
CS2D

Hajt5:
I replaced wrong game icon with official one (transparent background) and changed the name to "CS2D" because of [this](http://cs2d.com/faq.php?show=misc_name#misc_name).


rocco27:
Boht are good catch!
When I updated the icon I took the original one what I found on their official site and decreased the size to 16x16 and lost transparent background.
I also replaced the name to new one everywhere else except in lgsl_protocol.php file.
Thanks!

PR #206, opened 2017-08-22
getGameHomes_limit() Fix

Adjokip:
The join on home_ip_ports was causing servers wth no ports assigned to not appear under Administration->Game Servers

The data from that table wasn't used in relation to the getGameHomes_limit function - so it doesn't appear the join was ever necessary.


own3mall:
Excellent, thanks!  Yep, it doesn't need that join!

PR #210, opened 2017-09-14
Fix BUG in Game Server Monitor counter & Add pagination with search to groups page.

Obada8:
Hello all,

When i make search as a normal user for game server the count page make count for all game server in website.

[Screenshot](https://s26.postimg.org/mvyhr6bs9/chrome_2017-09-14_22-00-52.png)

So now it's fixed by this change.


own3mall:
I think the select statement isn't necessary.  How about using this in both places you added your changes:

```
AND home_id = '$search_field' AND home_cfg_id = '$home_cfg_id' 
```


Instead of this:
```
AND home_id IN (SELECT `home_id` FROM `".$this->table_prefix."server_homes` WHERE home_id = '.$search_field.' AND home_cfg_id = '$home_cfg_id' 
```

Obada8:
@own3mall 
I make test by replace this: `AND home_id = '$search_field' AND home_cfg_id = '$home_cfg_id' `
instead of this:
`AND home_id IN (SELECT `home_id` FROM `".$this->table_prefix."server_homes` WHERE home_id = '.$search_field.' AND home_cfg_id = '$home_cfg_id' 
`
But the result is not correct.

Obada8:
Ops, sorry for the groups pagination made under this title.
I was need to make it under separate title but i made mistake when i add it in GitHub, but i see that should work probably.

own3mall:
Explain why you have to do the in select statement then?  As far as I can tell, that statement does absolutely nothing.

Obada8:
@own3mall 
The main goal from
```AND home_id IN (SELECT `home_id` FROM `".$this->table_prefix."server_homes` WHERE home_id = '.$search_field.' AND home_cfg_id = '$home_cfg_id'```
Is to get all statements which defined in to equal with home_id associated with user.

If we put this:
`AND home_id = '$search_field' AND home_cfg_id = '$home_cfg_id'`
you'll see the home_id is separated from other clauses.

Obada8:
Oh sorry @own3mall , I just think that `IN` not work without `select` statement , sorry for that and sorry for bad English too.

Obada8:
Oh sorry @own3mall , I just think that `IN` not work without `select` statement , sorry for that and sorry for bad English too.

own3mall:
I think you can use "=", you just need to wrap your multi-conditions in parenthesis.

Obada8:
> I think you can use "=", you just need to wrap your multi-conditions in parenthesis.

Please can you give an example?

own3mall:
```AND (home_id = '$search_field' AND home_cfg_id = '$home_cfg_id') OR (something=something) OR```

etc

Obada8:
@own3mall I'm sorry for that i tried your method but not work, for me i don't see my edit is bad, you can accept and then edit what you think is better or close this pull.

own3mall:
OK for now... I guess.

PR #213, opened 2017-09-20
Added Ragnarok map to Ark maplist

Talung:
Added new Ragnarok map to the generated maplist for arkse_linux64.xml and arkse_win64.xml game modules.


rocco27:
Thank you!

PR #219, opened 2017-10-12
Allow all module menu names to be localized

Zorrototo:


Zorrototo:
Without that, menu names and administration submenus names have the name set in module.php and can not be localized.

PR #226, opened 2017-10-23
unused string

Zorrototo:
This was used back then when OGP was on sourceforge, it has no use now. Unless I'm wrong we can safely delete this from source lang file.


Zorrototo:
And it was old sourceforge version, it was removed here https://sourceforge.net/p/hldstart/svn/3242/

Zorrototo:
OK I commited to my master and I wanted to make a separate PR, but it added the commit the this PR. Whateverer, it just has two different things and title does not match now but these two commits are what I wanted to submit.

PR #228, opened 2017-10-24
Update max_user_amount & fix httport conflict

TAPL-93:


Zorrototo:
Insane but seems legit for a MTA-SA server https://wiki.multitheftauto.com/wiki/Multi_Theft_Auto

rocco27:
Even if this wiki says 4096 I believe to my test only, I tested every part of xml file when I updated it few weeks ago and 1024 was the MAXIMUM. This wiki is wrong...

TAPL-93:
You guys are using old version of the server. the maxplayer was increased to 4096 from years..
https://linux.mtasa.com

About the port, i already tested it, i created more than one server and only one can be started, the reset fail to start because the default port 22005 was binned to the first started server, also making it custom mean ports stealing .. that's why it preferred to set the httpport the same as serverport to prevent any conflict.

rocco27:
I tested latest version few weeks ago as I said I could not set 1025 maxplayers, game server denied higher setting. I saw 3 servers with 4096, 2000 and 1337 maxplayers, no more servers above 1024 limit. Those can be false, because those are hacked or modified somehow. It happens sometimes with some other games as well. Modders like to show higher slots as default, they think players.will join faster, etc. I can rwtest again, but I am sure it should be max. 1024 or less. 4096 also insane while every servers have only max. 50-500 players at the same time...so in this way makes no sense to have 4096 slots.

TAPL-93:
Your must be using Rsync which provide very old version, i guess 1.3?

Here the latest version:

==================================================================
= Multi Theft Auto: San Andreas v1.5.5
==================================================================
= Server name      : Test Server
= Server IP address: *****
= Server port      : 27015
= 
= Log file         : ..es/whmcs/65/mods/deathmatch/logs/server.log
= Maximum players  : 4096
= HTTP port        : 27015
= Voice Chat       : Disabled
= Bandwidth saving : Medium
==================================================================
[2017-10-24 09:09:49] Compacting accounts database 'internal.db' ...
[2017-10-24 09:09:49] Resources: 197 loaded, 0 failed
[2017-10-24 09:09:49] Compacting database 'registry.db' ...
[2017-10-24 09:09:49] Starting resources...
[2017-10-24 09:09:49] Server minclientversion is now 1.5.5-9.11725.0
[2017-10-24 09:09:49] Gamemode 'play' started.
[2017-10-24 09:09:49] Querying MTA master server... success! (Auto detected IP:*****)
[2017-10-24 09:09:49] Authorized serial account protection is enabled for the ACL group(s): `Admin`  See http://mtasa.com/authserial
[2017-10-24 09:09:49] Server started and is ready to accept connections!
[2017-10-24 09:09:49] To stop the server, type 'shutdown' or press Ctrl-C
[2017-10-24 09:09:49] Type 'help' for a list of commands.


Setting the default http port to always 22005 isn't the optimal solution, if you insist on the custom fields thing, you could make the default http port same as server port and at same time enable customization.

rocco27:
Who talks Rsync? It has what the owner upload. I was talking about the latest server files and I always grab it from the official site. I see your config. I will go and rebuild my test server because of you asap to prove my right...

Back to http port: do you understand what does it mean default setup? Yes we use default 22005 port for httpport and yes I gave the oppurtunity to change it in the custom fields which is important if you set up more than one game server. Thats the way what we should follow. If you insert any custom or personal things, then it will confuse others who read official tutorial and get different things from OGP!

TAPL-93:
https://github.com/OpenGamePanel/OGP-Website/pull/229#issuecomment-338957928

PR #229, opened 2017-10-24
fix httport conflict #2

TAPL-93:


rocco27:
Can you explain why are you destroying default configs for MTA?
Server port and http port are 2 different things in MTA. In the default server setup you have to set 2 different ports. You can set only one and it is working, but then it is a custom configuraton and we should follow default server setup.

TAPL-93:
> It's not possible to bind the same httpport to more than one server therefore it preferred to set the httpport the same as serverport to prevent any conflict

rocco27:
You can set http port for each servers in custom fields. So there will be no conflict.

TAPL-93:
https://github.com/multitheftauto/mtasa-blue/blob/master/Server/mods/deathmatch/mtaserver.conf

> but then it is a custom configuraton and we should follow always the default server setup!

Well, following the same logic, serverport is a custom configuration and we should follow always the default server setup!

< !-- This parameter specifies the UDP port on which the server will be accepting incoming player
         connections; default value: 22003. It is a required parameter. -->
< serverport>22003< /serverport> 


< !-- This parameter specifies the TCP port on which the server will be accepting incoming http
         connections. It can be set to the same value as < serverport>. It is a required parameter
         if <httpserver> is set to 1. -->
< httpport>22005< /httpport>

You can't just set httpport to always 22005 to any new created mta server. The custom fields doesn't have any check for duplication,  conflict will be there whatever it with server of the same game type or another. I'm still on my opinion it preferred to set the httpport the same as serverport to prevent any conflict and to be added "abuse".

rocco27:
Please check how xml files work first. Seems like you don't know this. For a parameter/option/setting we can set only one default value and not more. If it's editable the admin or the user can and maybe have change it. That's how httpport works in custom fields as well. I can say the default and if its a port you cuszomize if it is already used by another server which could be same type or different type of game server too.

TAPL-93:
Did you check the files changes of this pull request? it's to set single default value instead of multiple default value.

About the default ports, i will not argue anymore with this, here the thing: the default settings you are talking about are meant for a single game server which not been created with a panel, but when it comes to Open Game Panel the case here are different, the server will be created with a control panel and there may be multiple servers, if we apply your logic then whoever use this panel will have to bother himself to make sure there's no ports conflict. i myself hosted above 1000 MTA server since 2016, thankfully your logic was not applied at the time 🤣.

The only reason i'm creating this pull request is the generic users, otherwise i can always create my own custom xml.

I'm still on my opinion the optimal solution is to set the default httpport the same as serverport. There's no harm for setting httpport the same as serverport.

https://github.com/multitheftauto/mtasa-blue/blob/master/Server/mods/deathmatch/mtaserver.conf

< !-- This parameter specifies the TCP port on which the server will be accepting incoming http
connections. It can be set to the same value as < serverport>. It is a required parameter
if is set to 1. -->
< httpport>22005< /httpport>

own3mall:
Have you tested these configs?  Do you know for sure they work?  Just verifying, and then I'll merge.  

Thanks!

rocco27:
So I went to rebuild Test MTA Server to see am I able to host more than 1024 player slots.
I set 4096 maxplayers in the xml file. I choosed 4096 maxplayer before server start and guess what happend:

> ==================================================================
> 
> = Multi Theft Auto: San Andreas v1.5.5 [64 bit]
> 
> ==================================================================
> 
> = Server name      : club27 Test Server
> 
> = Server IP address: 188.165.225.155
> 
> = Server port      : 22003
> 
> = 
> 
> = Log file         : ..ser_Files/8/mods/deathmatch/logs/server.log
> 
> = Maximum players  : 1024
> 
> = HTTP port        : 22005
> 
> = Voice Chat       : Disabled
> 
> = Bandwidth saving : Medium
> 
> ==================================================================

rocco27:
I am testing. :D

rocco27:
So I went to rebuild Test MTA Server to see am I able to host more than 1024 player slots. I downloaded the server files from the official wiki again. This is the latest for Linux 64 bit. 

```
==================================================================

= Multi Theft Auto: San Andreas v1.5.5 [64 bit]

==================================================================

= Server name      : club27 Test Server

= Server IP address: 188.165.225.155

= Server port      : 22003

= 

= Log file         : ..ser_Files/8/mods/deathmatch/logs/server.log

= Maximum players  : 4096

= HTTP port        : 22005

= Voice Chat       : Disabled

= Bandwidth saving : Medium

==================================================================
```

Hmm, so this means this part of our discuss can be finished, you had right.

About httpport, I can just repeat myself, we should follow the default settings. If the server port is 22002 and httpport is 22005 as default then we do same in xml, this is fit for single server, and if you have multiple servers, you can just set different http port for each in custom fields menu.

As I mentioned before, the best solution could be if OGP can manage reserve ports in <replace_texts> and <custom_fields> as well, then we can set a math calc for http port e. g. httpport = server port +2 and insert in the <replace_texts> then it will write out to the config file. Until this I still recommend to do it manually.

TAPL-93:
@own3mall, I tested on latest MTA:SA server version and everything works.

Zorrototo:
I don't own this server, but I don't think binding two thing on same port is good idea. From what I understand, the custom field option was good before the previous accepted PR.

In this XML game port in OGP = server port = http port

Before you could set the http port separately, in custom fields, which was perfectly right I think, as admin you could decide which port to use, and you could allow user, or not, to change it by having proper permission when assigning the server to user/group. I don't understand the reasoning behind this with this XML changes.

TAPL-93:
Sorry about that, at the moment either merge #229 or revert #228.

rocco27:
Summary:
Thanks for previous maxplayers update. This "httpport" configuration will be revert back by me with another patch asap, so we keep the server and http ports separately.

own3mall:
@rocco27 - Please don't forget to get this done when you can.  Thanks guys!

PR #230, opened 2017-10-24
Update rsync_sites.list

FIREFOXCYBER:


Zorrototo:
Lol why adding your Rsync server in the middle? I would have it at the bottom personally, first because the other are the most trustable in my opinion (not that I don't trust you, I don't know you, but the other are owned by OGP devs), and because even alphabetically you're still at bottom, and also when we add something to a list, it usually at the bottom if there is no reason to add it elsewhere.

That doesn't make difference in the end but this is triggering my curiosity why you put your server there lol.

FIREFOXCYBER:
LoL N1 , You Can Add It EveryWhere I Dont Have Any Problem With This xD Trust ? i saw on rsync faq , u can send pm to add rsync to list , in faq just told about we need check ur server

Zorrototo:
Yeah you maybe misread what I wrote "**not** that I don't trust you, I don't know you, but the other are owned by OGP devs"

Just that adding it in the middle of the list is weird to me :D

rocco27:
Tested the link and it works. Thanks for setting up a new Rsync Server. The community will appreciate it.

FIREFOXCYBER:
http://rsync.rinetech.com , You Can See Games Now !

Zorrototo:
Just on a side note and maybe it is only personal preference, but I suggest that you hide Apache version and OS, for 'security' purpose in general. In case an exploit is discovered for your versions and/or OS, it makes it less obvious that you are vulnerable in case you're not up to date. Lot of tutorial on the internet for this it varies depending on OS version. Basically modify these lines to these values in the proper configuration file:
```
ServerTokens Prod
ServerSignature Off
```

FIREFOXCYBER:
Ok, Thanks for Suggestion!

PR #233, opened 2017-10-27
Make Extras modules update result more readable

Zorrototo:
separate each group (overwritten, new, not overwritten, and module version), with two lines space to make it more readable instead of having the next group right after the last filename of previous group.

before: https://www.dropbox.com/s/4pv01v2sen1hekd/Capture%20d%27%C3%A9cran%202017-10-27%2020.37.19.png?dl=0
now: https://www.dropbox.com/s/xxi4zi4c7mx2hgc/Capture%20d%27%C3%A9cran%202017-10-27%2020.53.37.png?dl=0


rocco27:
Looks much better. 👍

PR #245, opened 2017-11-05
Monthly update #1

rocco27:
1, New game configs for:
- Mafia 2 Online

2, Game config updates for (few small changes and added "lock_files" and "configuration_files"):
- 7 Days to Die
- Aliens vs Predator
- Empyrion - Galactic Survival
- Homefront
- Hurtworld
- Squad

3, TeamSpeak 3 PHP Framework update from v1.1.25 to v1.1.27


modules/config_games/server_configs/hurtworld_linux32.xml
```diff
@@ -25,4 +25,20 @@
       <installer_login>anonymous</installer_login>
     </mod>
   </mods>
+  <post_install>
```
Zorrototo: Just my two cents, but what if this file is updated? Wouldn't it better to copy it everytime server is updated, just in case?

PR #246, opened 2017-11-06
Update TS3Admin.php

WSG-ViViV:


Zorrototo:
localization are done on Transifex and linked on the forum. I'll correct it.

PR #250, opened 2017-11-13
fix invalid schema when using occurrence

GMckeezy:


own3mall:
Occurrence should only be used for text replacements?  How does this fix anything?  In fact, it breaks what I did.

own3mall:
Added your change and updated some other things too in a different commit.  Thanks.

PR #272, opened 2017-11-26
Theme Include Consitency for Modules

oNdsen:
Small Loop which checks Theme specific Files
A css or js File in Root Dir of a Module could be used for all Subpages instead of copy it.
Should not affect already exist Theme Files, because it works recursive backwards (see Array).

### DIFF
--- a/includes/view.php
+++ b/includes/view.php
@@ -121,10 +121,18 @@ class OGPView {
 		$module = isset($_GET['m']) ? $_GET['m'] : "";
 		$subpage = isset($_GET['p']) ? $_GET['p'] : $module;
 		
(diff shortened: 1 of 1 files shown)


own3mall:
What's the purpose of this commit?  What does it do that the code didn't do before?

oNdsen:
it checks for a "root" file (css or js) and includes it instead of the default theme file in specific module.
with this option a theme dev is able to create only one css and js file for all subpages at a module.
example:
module: foo
subpage1: bar
subpage2: bar2
subpage3: bar3
theme include file: /modules/foo/foo.css
now this file will be included on all subpages (bar1, bar2, bar3). BUT if bar2.css exists, on this subpage bar2.css will be included instead of foo.css

PR #279, opened 2017-11-28
Language Support in JS

own3mall:


own3mall:
@Adjokip - Can you think of any security issues this may cause?  I can't...

own3mall:
@Adjokip , I merged it, but I'd still like to know your opinion on any potential security issues that might be opened.

Adjokip:
I think the only issue would be is if the language constant contained some malicious data. But if everything is reviewed prior to pushing, it should be fine. htmlentities and/or striptags would resolve that possible issue.

Another possible issue would be is anywhere with define and a password or sensitive data as that would be exposed to the user. But I don't believe that happens anywhere.

own3mall:
Thanks @Adjokip, I thought of those use cases as well.  It looks like it has to be htmlentities anyways to get the French characters right, so that's done now too ;)  -  If you can think of any other risk, please let me know :+1:

PR #285, opened 2017-11-30
Massive Database Changes

own3mall:
Everything should work... I hope.


own3mall:
@Adjokip , any thoughts?  This fixes the double escape issue.

PR #286, opened 2017-11-30
Sanitize some inline queries

Adjokip:
With the large change to the database_mysqli.php file (#285) which removed escaping on every `_POST`, `_GET` and `_REQUEST` variable due to some double escaping, there was some in-line queries ran which were now vulnerable.

The change to English/global.php is because the constants are already defined.

`addonsmanager/addons_manager.php` appears as a large diff, I believe that's because of the line endings on the current version on Github - see: https://github.com/OpenGamePanel/OGP-Website/blob/master/modules/addonsmanager/addons_manager.php

### DIFF
--- a/lang/English/global.php
+++ b/lang/English/global.php
@@ -130,8 +130,6 @@ define('no_results_found', "No search results found for %s");
 define('tickets', "Support Tickets");
 define('news', "News");
 define('admin_news', "News Admin");
-define('rcon', "RCON");
-define('support', "Support");
 define('util', "Utilities");
 define('fast_download', "Fast Download");
 define('fd_user', "Fast Download");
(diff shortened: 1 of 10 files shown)


own3mall:
Looks good, but I don't want the constants removed from global.  They're probably there for a reason even if the modules may conflict individually.  I'll find a better way to handle this use case.

Adjokip:
The `support` and `rcon` constants were removed as they're already defined in the same file - on lines 34 and 117.

Zorrototo:
The constants in global.php are here to be able to localize menu names for modules. I added them https://github.com/OpenGamePanel/OGP-Website/pull/219 but I overlooked as they were already in the file when I did it lines 34 and 117. weird I didn't see/find them.

PR #289, opened 2017-12-02
Monthly update #1

rocco27:
Game config update for:
- Multi Theft Auto xml update

Changes for the Custom Fields menu:
- Removes unnecessery buttons its page
- Renamed the string and its texts

### DIFF
--- a/lang/English/modules/gamemanager.php
+++ b/lang/English/modules/gamemanager.php
@@ -204,5 +204,5 @@ define('cid', "CID");
 define('phan', "Phantom");
 define('sec', "Seconds");
 define('unknown_rsync_mirror', "You attempted to start an update from a mirror which doesn't exist.");
-define('go_to_custom_fields', "Go to Custom Fields");
+define('custom_fields', "Custom Fields");
 ?>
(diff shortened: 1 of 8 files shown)


lamjed001:
and don't forget to add new custom fields

`    <field key="Owner Email" type="text">
      <default>owner_email_address</default>
      <default_value></default_value>
      <filepath>mods/deathmatch/mtaserver.conf</filepath>
      <options>tags</options>
      <desc>This parameter specifies the contact email addresses for the owner(s) of this server.</desc>
    </field>
    <field key="Voice" type="select">
      <default>voice</default>
      <option value="1">Enable</option>
      <option value="0">Disable</option>
      <default_value>0</default_value>
      <filepath>mods/deathmatch/mtaserver.conf</filepath>
      <options>tags</options>
      <desc>This parameter specifies whether or not to enable player voice chat in-game.</desc>
    </field>`

rocco27:
Sorry, but I don't feel those settings must be in the Custom Fields, however users always can set in the config file by using easy way (Edit configuration files) or very difficult way (open file manually by using File Manager or FTP client).

If you feel you must have, you can add to your xml and insert the file to the blacklist, so Update module won't touch your customized file anymore.

PR #301, opened 2017-12-14
Update DayZ config to not use hardcoded values for mods and fix multi-instance config

PrimusNZ:
- Added the 5 most common DayZ flavours as a dropdown instead of having to edit the template for different DayZ server types
- Add port as 2 or more instances on the same machine without this parameter, instances after the first wont start.
- Fix config path as didnt work (Didnt need escaping)
- All Dayz instances use `dayz` folder for config


PrimusNZ:
This is a fix to my previous version of the template which didnt work too well

PR #303, opened 2017-12-17
Update MySQL users to use all hosts

PrimusNZ:
Changes to the mysql admin  module to create database users with host being allow all hosts.

Up to user to configure MySQL to listen on `0.0.0.0` (Or whatever IP they prefer) and ensure correct firewalling.

Didnt see a migration path for existing users, Existing users will still work, but host should be manually updated to `%`


own3mall:
Thanks.

krille02:
Hi how to change mysql for not accessing my sql files with this HeidiSQL. it says that mysql server on my ip number. i am running linux server

PR #308, opened 2017-12-31
added get_lang function where it was missing

Zorrototo:


Zorrototo:
Also noticed that some print_success or print_failure were not localized at all they are hardcoded messages.

PR #309, opened 2018-01-01
get_lang added to rcon.php and update_server_manual.php

Zorrototo:


Zorrototo:
in mini_start.php get_lang_f is used instead of get_lang in many places when I think it shouldn't. need to check this later

PR #311, opened 2018-01-03
Remove rcon2 as control protocol for DayZ Mod

PrimusNZ:
Doesnt seem to work, prevents ogp_api from shutting down the server

rcon2 doesnt seem to be a protocol supported by A2 (I suspect Arma use their own RCON protocol)

### DIFF
--- a/modules/config_games/cli-params.php
+++ b/modules/config_games/cli-params.php
@@ -168,6 +168,7 @@ function exec_ogp_module() {
 	   <option value="">None</option>
 	   <option value="rcon">HL/Quake/CoD Support</option>
 	   <option value="rcon2">HL2/Source Support</option>
+	   <option value="armabe">Arma 2 BattlEye Support</option>
 	  </select>
 	 </td>
 	</tr>
--- a/modules/config_games/schema_server_config.xml
+++ b/modules/config_games/schema_server_config.xml
@@ -63,6 +63,7 @@
       <xs:enumeration value="rcon" /><!-- HL, Q1/2/3 -->
       <xs:enumeration value="rcon2" /><!-- HL2(source) -->
       <xs:enumeration value="lcon" /><!-- legacy console -->
+      <xs:enumeration value="armabe" /><!-- Arma 2 BattlEye -->
       <xs:minLength value="1" />
     </xs:restriction>
   </xs:simpleType>
--- a/modules/config_games/server_configs/arma2_win32.xml
+++ b/modules/config_games/server_configs/arma2_win32.xml
@@ -11,7 +11,7 @@
   </cli_params>
 <console_log>profile/server_console.log</console_log>
 <max_user_amount>64</max_user_amount>
-<control_protocol>rcon2</control_protocol>
+<control_protocol>armabe</control_protocol>
 <mods>
  <mod key='arma2'>
   <name>none</name>
--- a/modules/config_games/server_configs/arma2co_win32.xml
+++ b/modules/config_games/server_configs/arma2co_win32.xml
@@ -11,7 +11,7 @@
   </cli_params>
 <console_log>profile/server_console.log</console_log>
 <max_user_amount>64</max_user_amount>
-<control_protocol>rcon2</control_protocol>
+<control_protocol>armabe</control_protocol>
 <mods>
  <mod key='arma2co'>
   <name>none</name>
--- a/modules/config_games/server_configs/arma2oa_win32.xml
+++ b/modules/config_games/server_configs/arma2oa_win32.xml
@@ -11,7 +11,7 @@
   </cli_params>
 <console_log>profile/server_console.log</console_log>
 <max_user_amount>64</max_user_amount>
-<control_protocol>rcon2</control_protocol>
+<control_protocol>armabe</control_protocol>
 <mods>
  <mod key='arma2oa'>
   <name>none</name>
--- a/modules/config_games/server_configs/dayz_arma2co_linux.xml
+++ b/modules/config_games/server_configs/dayz_arma2co_linux.xml
@@ -10,7 +10,7 @@
   <cli_param id="PORT" cli_string="-port=" />
  </cli_params>
  <console_log>cfgdayz/arma2oaserver.RPT</console_log>
- <control_protocol>rcon2</control_protocol>
+ <control_protocol>armabe</control_protocol>
  <mods>
   <mod key='dayzmod'>
    <name>none</name>
--- a/modules/config_games/server_configs/dayz_arma2co_win32.xml
+++ b/modules/config_games/server_configs/dayz_arma2co_win32.xml
@@ -9,7 +9,7 @@
   <cli_param id="PORT" cli_string="-port=" />
  </cli_params>
  <console_log>cfgdayz/arma2oaserver.RPT</console_log>
- <control_protocol>rcon2</control_protocol>
+ <control_protocol>armabe</control_protocol>
  <mods>
   <mod key='dayzmod'>
    <name>none</name>
--- a/modules/config_games/server_configs/dayz_arma2oa_win32.xml
+++ b/modules/config_games/server_configs/dayz_arma2oa_win32.xml
@@ -15,7 +15,6 @@
  <cli_allow_chars>;</cli_allow_chars>
  <console_log>profile/arma2oaserver.RPT</console_log>
  <max_user_amount>64</max_user_amount>
- <control_protocol>rcon2</control_protocol>
  <mods>
   <mod key='dayzmod'>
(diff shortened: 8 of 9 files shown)


Zorrototo:
you suppose or this is something sure?

PrimusNZ:
Im sure. Arma 2 an Arma 3 use its wn version of RCON via BattlEye

I am working on adding BattlEye RCON support to OGP

PrimusNZ:
I have added Arma BattlEye support. Pull requests in agents to support

own3mall:
Why would it be trying to shutdown the server using rcon?  If it fails to shutdown using rcon, the process should still terminate.

PrimusNZ:
Not sure, this is the behaviour I was experiencing:
https://www.opengamepanel.org/forum/viewthread.php?thread_id=5904

Removing rcon2 solved the issue, but just killing an Arma server is bad (Game progress doesn't save)
Adding the Arma Battleye support like I have causes the server to be shut down gracefully is is much better and preferred

own3mall:
I see.  Remove the "sleep 5" lines from the agents, and I will merge this pull request.  

Those lines will freeze the agent and won't actually do anything of use.

PrimusNZ:
They actually do, in testing these changes, I noticed that after sending the shutdown command, immediately after it, the agent would kill the server (Why it wasn't doing this with rcon2 even though it didn't work I don't know), but it wasn't allowing the server to save. The sleep was enough to give the server time to shut down before the OGP agent forcefully killed it.

Is there a way I can suppress the auto-kill  without using sleep to wait for the server to shut down gracefully?

own3mall:
It should have been killing the server before.  If it wasn't, there might be a bug.  I'll have to look into this when I'm back from vacation.  

I really don't like the idea of delaying the shutdown of any game process.  Is there a way to force a save command via rcon before the server is shut down?  It never saves automatically?  Why does Arma work this way, and what state is being saved server-side?  That's crazy...

own3mall:
If your PERL extension for ARMA has a return code that signals the server will shut down properly, then I would be OK with a skip variable for the forceful termination, but it should have still happened previously in your case with rcon2.

own3mall:
Also, did you write that module or find somewhere?  :P

PrimusNZ:
The perl module.does have a return code yes, you can see this being checked in the agent code I added for the shutdown command. The sleep only applies if a successful result was returned, otherwise forcefully termination.

Yeah Arma is weird, mostly cause it's got so many mod game modes (Exile, Epoch, DayZ epoch, DayzMod)
Most saving occurs through the server running process (Timed or event based), but the server does a final save on shutdown(If the mod is set up to do that). States being saved are player stats (Weapon, gear, health etc), base objects, vehicle positions etc. Every mod is different in how they handle their saving

I found it as a script, got it working, then cannibalised the KKRCON/HL2.pm module and got it working (I have very little perl experience, enough to get by)

own3mall:
What does your ogp_agent log say (before you use your armabe changes) when rcon2 doesn't stop it?

It should say something like:

> logger "Failed to send rcon quit. Stopping server with kill command.";
> logger "Stopped process with pid $pid successfully using kill 15.";
> logger "Stopped process with pid $pid successfully using kill 9.";

I reviewed the code, and the agent should terminate the server regardless if rcon doesn't stop it... so I'd like to look into that issue a little as to why it wasn't stopping your arma server forcefully.

As for your changes, I'm going to merge them for now, but I'm going to ask you to test a modified version for me that is going to allow a sleep time of up to 5 seconds, but may not have to wait that long... I don't think the code will block based on the fact that I'm guessing every call to the agent is threaded in a way since Frontier::Daemon::Forking is technically a web server...

own3mall:
Could you please try this version of the agent and see if it gives your server enough time to shutdown using your new armabe rcon protocol without using a fixed sleep?


[ogp_agent.pl.zip](https://github.com/OpenGamePanel/OGP-Website/files/1636838/ogp_agent.pl.zip)

PrimusNZ:
Will do, I'll have to check later on this evening

PrimusNZ:
@own3mall Is this zip the windows or linux agent version. (I only have a windows agent installed)

own3mall:
It's Linux... let me upload a test Windows one.

own3mall:
[ogp_agent.pl_windows.zip](https://github.com/OpenGamePanel/OGP-Website/files/1637022/ogp_agent.pl_windows.zip)

Hopefully there aren't any errors in it... can't test for a quick syntax check easily at the moment since I primarily run Linux.  But, you can get the idea by what I'm trying... let me know if it works please :+1:

PrimusNZ:
@own3mall I'll be checking tonight, Sorry went away for the long weekend

own3mall:
@PrimusNZ I merged these changes in by accident anyways... guess we'll find out shortly if it works :+1:

PrimusNZ:
I tried the other day, but I couldnt get the updater to stop overwriting the file, so I gave up for the night
I can update and test later this evening since its been merged. The code looks sound though

PrimusNZ:
Okay, doesnt seem to be working

Log File: [https://pastebin.com/R24DnUtL](url)
Web interface gives me: `The Agent that is controlling this server is offline`
Agent is running, doesnt appear to crash or anything

PrimusNZ:
Played around with the ogp_agent.pl a little adding some logging.

Seems `stop_server_without_decrypt` is being called twice ? First time with control protocol and control password, second time without control password. Should that method be being called twice ?

Going to add logging everywhere its used in the file and determine where its being triggered

own3mall:
I'm guessing it has to do with the agent remote time out setting, but I'll revisit it again soon.

own3mall:
Yes, increase your agent timeout to 10 seconds, and then try it again.  You'll want to use a larger seconds value than what the agent uses for a soft-shutdown time for game servers that support a protocol.  

You can set the value the agent uses in the Agent Install Directory --> Cfg --> Preferences.pm file.  Simply add the following to the end before the closing parenthesis:

```
protocol_shutdown_waittime => '5',
```

So for example, mine looks like this:

```
%Cfg::Preferences = (
	screen_log_local => '1',
	delete_logs_after => '3',
	ogp_manages_ftp => '1',
	ftp_method => 'EHCP',
	ogp_autorestart_server => '1',
	protocol_shutdown_waittime => '5',
);
```

If you change your settings, restart the agent service:
```
sudo service ogp_agent restart
```

own3mall:
I've documented this as well here:

https://opengamepanel.org/forum/viewthread.php?thread_id=5996

own3mall:
Shouldn't your changes apply to the Linux version of Arma2 as well?

PrimusNZ:
@own3mall In theory yes since the BattlEye protocol should be the same, But as i dont (And have never) run A2 on linux, I could be wrong.

PrimusNZ:
Nope still experiencing a double event. Panel timeout to the agent is set at 60, `protocol_shutdown_waittime` is set to 30 and im still seeing a kill event directly after the shutdown event.
Almost as if the agent is returning a failed on the first try.

PrimusNZ:
I added logging to the shutdown function and it seems its not even getting to the wait loop.
I added logging throughout the shutdown function to see if it was quitting, see my pastebin: 
[https://pastebin.com/jf06e2Ug](https://pastebin.com/jf06e2Ug)

Seems to get to the `my @server_pids = get_home_pids($home_id);` line and stops. the preUsedToStop log-line never makes it to the log file

PrimusNZ:
Actually you know what? `get_home_pids` doesnt exist in the windows agent.
[Windows Agent](https://github.com/OpenGamePanel/OGP-Agent-Windows/search?utf8=%E2%9C%93&q=get_home_pids&type=)
[LInux Agent](https://github.com/OpenGamePanel/OGP-Agent-Linux/search?utf8=%E2%9C%93&q=get_home_pids&type=)

own3mall:
Yep, you're right.  Somehow I thought it was there when I implemented it... anyways, I'll fix that for Windows.

own3mall:
It should be fixed in the latest version.

I think we should make the control protocol change in the Linux config files as well for Arma2.

PrimusNZ:
I concur. Want me to do it and submit a PR or do you wanna just go ahead and make the change ?

PrimusNZ:
Just tested this update/fix. Working beautifully now.

own3mall:
I closed my virtual machine already.  If you could create a pull-request, that would be great.  We want to stay consistent in the game server XML.

PrimusNZ:
Created: [Update Arma 2 Linux Templates to use armabe control protocol](https://github.com/OpenGamePanel/OGP-Website/pull/317)

own3mall:
Thanks!

PR #316, opened 2018-01-30
FiveM XML

Adjokip:
Requested here: https://www.opengamepanel.org/forum/viewthread.php?thread_id=5747

A key is required to run servers, which can be generated at https://keymaster.fivem.net/

The IP / Port the server binds to is specified in the server.cfg file. So cfg_text_replace was changed so the IP / Port may be retrieved via the ip_port key in the appropriate format (IP:Port - I don't think that was available for replace_texts before)

The post install script handles the installation of the server. It'll pull the newest version of the server for the appropriate platform.


scarecr0w12:
This should really be removed, as it is against Rockstar games TOS to run or host private servers.

> There are only two authorized Grand Theft Auto online services – the Rockstar Games’ hosted Grand Theft Auto IV multiplayer feature and Grand Theft Auto V’s Grand Theft Auto Online. Any other online Grand Theft Auto service is unauthorized and your use would violate the license for your software. If you develop software for or host unauthorized servers you are at risk of legal action by Take Two Interactive. Players who use unauthorized servers may be banned from Rockstar Games services.

https://support.rockstargames.com/hc/en-us/articles/115004754228

own3mall:
@scarecr0w12 The legality of games or their mods is not up to us to decide.  All that is provided is an XML file.  Use at your own risk.

PR #321, opened 2018-02-11
Modified form table class to match the new OGP lang format

Zorrototo:
This way when it creates table, and there is a matching string with "_info" at the end, it creates the help icons back like before on the settings page and many other Panel pages.

### DIFF
--- a/includes/form_table_class.php
+++ b/includes/form_table_class.php
@@ -124,7 +124,7 @@ class FormTable {
 			print_failure(get_lang_f('invalid_setting_type',$type));
 		}
 
-		if ( defined($name."_info") )
+		if ( defined("OGP_LANG_".$name."_info") )
 		{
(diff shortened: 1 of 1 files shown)


Zorrototo:
I think we may make this better, because there are still inconsistencies between language file, and the PHP Panel files where we'll have different cases: "OGP_LANG_strinxxx" or simply "stringxxx"

I guess we could add a fail safe thing like:

`if ( defined("OGP_LANG_".$name."_info") || defined($name."_info"))`

This is for the case where the lang file does not have the format with "OGP_LANG_"

own3mall:
It should be using the "OGP_LANG_" prefix now anyways.  Is there somewhere where it's not?  The get_lang function allows the old format just so I wouldn't have to update EVERYTHING, but for the _info strings, they should be using the OGP_LANG prefix.

Zorrototo:
It may be possible that extra modules don't have the OGL_LANG thing I don't know.

PR #350, opened 2018-06-05
Update: theforest_win32.xml

jonathanhagler:
-- Changed the %SAVE_PATH% option under <cli_template> to use a backslash instead of a forward slash to match Windows path syntax and correct error in save file location
-- Added the following server options now available in v1.07
---- difficulty (Peaceful, Normal, Hard)
---- veganmode (no enemies)
---- vegetarianmode (no enemies during daytime)
---- resetholesmode (Reset all existing floor holes when loading a save)
---- treeregrowmode (Enable Tree regrowth when sleeping)
---- nobuildingdestruction (Enemies and players cannot destroy buildings)
---- allowenemiescreative (Enable enemy spawn in creative games)
---- allowcheats (Allow clients to use the built in development console)
---- serverpassword (Password protection for game)

### DIFF
--- a/modules/config_games/server_configs/theforest_win32.xml
+++ b/modules/config_games/server_configs/theforest_win32.xml
@@ -5,7 +5,7 @@
   <installer>steamcmd</installer>
   <game_name>The Forest</game_name>
   <server_exec_name>TheForestDedicatedServer.exe</server_exec_name>
-  <cli_template>-batchmode %IP% %PORT% %QUERY_PORT% %STEAM_PORT% %HOSTNAME% %PLAYERS% %CONTROL_PASSWORD% %IT% %SLOT% %SSA% %VAC% %SAVEINTERNAL% %SAVE_PATH%/ %LOG% -nographics</cli_template>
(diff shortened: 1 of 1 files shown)


own3mall:
Pending @rocco27 review

rocco27:
Thank you for your update. Summing:

1) Foward slash work fine, tested on both OS before released xml. I prefered foward slash, because Agent is running under Linux on Windows too. :)

2) All new server settings are game setttings which can be configured by editing the config file manually in File Manager/FTP or using Editconfig files module.

If you want I can explain many reasons ehy don't like the idea...

I do not prefer to use OGP Pabel to set any game server settings, only if it is necesswry because works only in command line.

jonathanhagler:
@rocco27 Thanks for this. I apologize for the wasted time. I think I misunderstood the way some of the systems work. 

1. That makes sense. I was only thinking about it from the Windows agent point of view. Thanks for helping me understand.

2. I must have something wrong with my install. Editing the config using the editconfig module did not yield expected results for me. When I observed the command line passed to the dedicated server application it did not appear to reference the config file. And when reading the game server documentation it appeared that you could use passed command line arguments OR the config file. I was probably just doing something wrong.

My apologies for the waste of time reviewing the xml.

rocco27:
The config file path is not working well yet in ds, so that's why OGP cannot manage config file yet. Game server generates it a hidden and wrong path. Everyone waits for Forest devs for fix this small, but serius bug for 4 weeks now. Sadly serms like they are just busy with VR function only.

jonathanhagler:
Sounds about right. I saw some complaints on other forums about the CLI argument being broken.

rocco27:
Fixed my comment. Also recommend you to check Forest thread on forum to continue discuss.

PR #365, opened 2018-07-11
Update README.md

carterkelly9:
A few small grammar fixes!


Zorrototo:
You're doing it wrong. 1800 files changed? no.

Zorrototo:
Please use the correct Branch.. Obviously it is called MASTER not "Adjokip-patch-1" from 8 months ago (by the way this branch could be deleted @Adjokip if you're not using it anymore, even if you still contribute I don't think you use it and it was a one time use).

carterkelly9:
My bad, don't know how I ended up doing that haha, will fix!

PR #372, opened 2018-07-24
Center new API setting link

Zorrototo:


own3mall:
It would be better to use a CSS class if possible for the alignment so that themes could change it if they wanted... I'm not the best about this either, but that is better practice.

Also, can you post a screenshot showing before and after?

Zorrototo:
Here is current: https://www.dropbox.com/s/8b1cen8hbuegwji/Capture%20d%27%C3%A9cran%202018-07-24%2018.23.19.png?dl=0
Here is with centered DIV: https://www.dropbox.com/s/u21d7yoth5m0kw8/Capture%20d%27%C3%A9cran%202018-07-24%2018.22.44.png?dl=0
Here is now with H4 tag: https://www.dropbox.com/s/siwxg8holh5p9er/Capture%20d%27%C3%A9cran%202018-07-24%2018.28.58.png?dl=0

own3mall:
Looks good.  Thanks :+1:

PR #376, opened 2018-07-29
Removing DSi administration icon from OGP Master

Zorrototo:
in order to add it back to the DSi module repo


Zorrototo:
will work with https://github.com/OpenGamePanel/Module-DSi/pull/17

PR #377, opened 2018-07-30
Consistency

Zorrototo:


DieFeM:
The token has 64 characters length, what do you mean by "Consistency", I don't get it, sorry.

Zorrototo:
It is the length of the input, all others are 50 pixels wide.

PR #383, opened 2018-08-08
Create Arma3_headless_client_linux32.xml

tfranz42:
Arma3 Headless client XML for linux32-bit rough first pass.


rocco27:
Please use low capitals only in the file names
Also would be nice to add win32 version too. Then we support all supported OS. Talk more soon on forum.

tfranz42:
I'll go ahead and close this request then and hold off till after we talk on the forums. Sorry about derping on the capitalization. I'll work on the 32bit version as well.

PR #403, opened 2018-12-08
Update spigotmc_linux.xml

MrDKGE:
-XX:+UseConcMarkSweepGC will replace the old -Xincgc .
-Xincgc will result in java printing out the error : Could not create the java virtual machine
This fix will work on openjdk-11-jdk and openjdk-8-jdk for Linux

Tested on Ubuntu 18.04/16.04 with both openjdk 11/8 .

### DIFF
--- a/modules/config_games/server_configs/spigotmc_linux.xml
+++ b/modules/config_games/server_configs/spigotmc_linux.xml
@@ -4,7 +4,7 @@
   <gameq_query_name>minecraft</gameq_query_name>
   <game_name>Spigot Server</game_name>
   <server_exec_name>spigot_run</server_exec_name>
-  <cli_template>-Xincgc %XMX% -jar spigot.jar nogui</cli_template>
+  <cli_template>-XX:+UseConcMarkSweepGC %XMX% -jar spigot.jar nogui</cli_template>
   <max_user_amount>32</max_user_amount>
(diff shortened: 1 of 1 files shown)


Zorrototo:
Any documentation? Any reason why we should change this? I mean official documentation and explanations.

MrDKGE:
Sorry for the lack of Documentation.
Here is the installation steps from spigot them self https://www.spigotmc.org/wiki/spigot-installation/
If i got it correctly for java versions 8 and up the command is no longer supported.
https://openjdk.java.net/jeps/214
http://openjdk.java.net/jeps/173
More Documentation needed please let me know i can run test with screenshots as well to see the result when using -Xincgc on java 8 and up.

MrDKGE:
Ignore the last 2 commits about rsync i didn't know it would automatically sync it to this request. i will remove the changes.

Zorrototo:
Yeah you need to create 'Branches' to separate the commits :D it is confusing to use GitHub sometimes :D

I'll have some read thanks for links. If this is the new official way, I guess we could go for it then, but I'll wait for people who deal with this type of java servers to be sure.

Zorrototo:
@rocco27 I just noticed this is still pending. Any though about accepting or not this PR?

PR #405, opened 2019-01-09
Add Miscreated Dedicated Server - Win64

PrimusNZ:
Changes to add support for the recently released Miscreated Server files
These appear to be win64 only so only created win64 template


Zorrototo:
Can't test as I do not use Windows OGP, but seems right from first read.

Zorrototo:
I'm going to test this XML in Wine soon as apparently Linux servers are not going to happen on Cry Engine games, this is the only possibility.

Zorrototo:
I noticed a problem already, you proobably have your personnal steam account in your panel settings and can download, but I can't download it, need to use anonymous login in XML.

Still don't know about the map yet.

PR #412, opened 2019-01-27
Added Europa to the map list and types of map.

Talung:
Added Europa to the list and also added in type of map.  ie. Survival or Creative.

### DIFF
--- a/modules/config_games/server_configs/stationeers_linux64.xml
+++ b/modules/config_games/server_configs/stationeers_linux64.xml
@@ -45,10 +45,11 @@
       <desc>Interval in Seconds for save</desc>
     </param>
     <param key="-worldtype=" type="select" id="MAP">      
-      <option value="Moon">Moon</option>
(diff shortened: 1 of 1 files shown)


rocco27:
Hi there! I just did a quick look to the dedicated server configuration and it shows only:
> -worldtype | string | Worldtype(Space, Mars, Terrain)
Is it possible the wiki is outdated? I don't know this game server...
Link: [https://stationeers-wiki.com/Dedicated_Server_Guide](Dedicated Server Guide)

Talung:
Hi rocco27, 

Yeah that Wiki is not up to date.  They added things like modding etc. but haven't updated the map list at all.  I got the maplist from the ingame menu.  Tried them out and they all work properly.  I think the "-worldtype" serves more as an "example" and not an exact list.  Nature of Early Access.

rocco27:
Okey, thank you for your answer and changes.

PR #434, opened 2019-06-05
Minor TS3 issues fixed

Zorrototo:
Two minor issues fixed for TS3 in OGP:
- the 'map image' used in game monitor was the TS2 one
- the icon used in the transition page when starting a TS3 server was not the TS3 server icon as it was failing to find it


Zorrototo:
If people can test and verify this PR, that would be great. My OGP installation is kinda modified and I would like someone else to check I didn't break their OGP :+1:

rocco27:
I tested and both works. 👍

PR #462, opened 2020-01-07
Create blackmesa_source_server.xml

OTD2019:


Zorrototo:
Please change the game_name because it is obviously a server, no need to extend the name with " - Server", also the game name is "Black Mesa" not "BlackMesa".

Also please write everything in ENGLISH, I see German in your XML file from line 42 to 45.
Anyway, I guess this whole section could be removed and then you could directly add the startup parameter to the command line, because if this XML is for Black Mesa, I find it weird that there is a server start parameter to select if server starts on Black Mesa, or not.
Maybe you have an actual reason to do that?

You have an empty section, the Custom Fields, remove that.

Maybe I'm too demanding, but I would also create a new entry for this in the [LGSL protocol file](https://github.com/OpenGamePanel/OGP-Website/blob/master/protocol/lgsl/lgsl_protocol.php) for the game to have its own icon and own game name (for example in the extra LGSL module or DSI module) instead of the Half Life ones. Here look [an example for Post Scriptum](https://github.com/OpenGamePanel/OGP-Website/commit/4bf7f74dfa52490ff5a89ba9510a23daec9d6ba6#diff-d2917a1bba1d70f9b8c27ed9c070a6cd), do exactly the same as on lines 114 (`blackmesa` / `Black Mesa`), 278 (`blackmesa` / `05`) and 439 (`blackmesa` / `steam://connect/{IP}:{S_PORT}`). You can ignore the Squad lines of course and I think the part for the ports information though it isn't required for source games I think). Please know your alphabet to properly place your new lines at the proper order in the file, this shouldn't be complicated. when your protocol file is done, change your XML to use this 'new' LGSL protocol called "blackmesa" (which is, like many game server, some kind of redirection to the Source protocol, the number "05").

And add an icon then too with same file name as your protocol name (`blackmesa.png`).

Zorrototo:
I'll redo it all next week too.. Sad that people don't even reply and try.. always need to do it all for them..

PR #469, opened 2020-02-19
Fixed server info/controls not showing when expanding via address/owner column

xRJx:
I encountered an issue on Ubuntu 18.04 panel version d0c4f2b386cdc1e1172f899f56f5567eb02ff2ef (tested on Chrome, Firefox, and Internet Explorer) where the server's info and start/stop controls wouldn't show up when expanding the collapsible accordion by clicking on the address/owner columns (the info and controls would show up when clicking on the server name column). This change fixes that issue.

### DIFF
--- a/modules/gamemanager/server_monitor.php
+++ b/modules/gamemanager/server_monitor.php
@@ -494,7 +494,7 @@ function exec_ogp_module() {
 			// Template
 			@$first = "<tr class='maintr$trclass'>";
 				$first .= "<td class='collapsible sortHandle' data-status='$status' data-pos='$pos'><span class='hidden'>$order</span>" . "<img src='" . check_theme_image("images/$status.png") . "' />" . "</td>";
-				$first .= "<td class='collapsible sortHandle'>" . "<span class='hidden'>$mod</span><img src='$icon_path' />" . "</td>";
+				$first .= "<td class='collapsible sortHandle' data-status='$status' data-pos='$pos'>" . "<span class='hidden'>$mod</span><img src='$icon_path' />" . "</td>";
 				$first .= "<td class='collapsible serverId hide sortHandle'>" . $server_home["home_id"] . "</td>";
 				$first .= "<td class='collapsible serverName ignoreSortable' data-status='$status' data-pos='$pos'><b>" . htmlentities($server_home['home_name']) . "</b>$mod_name</td>";
 				$first .= "<td class='collapsible serverIPAddress ignoreSortable' data-status='$status' data-pos='$pos'>" . $address . "</td>";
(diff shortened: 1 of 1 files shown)


own3mall:
The old code is more desirable and was intentionally done for the re-ordering functionality that was introduced in the past.

xRJx:
My apologies, I don't think I was clear enough as to what I was talking about. The issue isn't encountered when clicking on the headers of the columns (which would re-order things), but rather when clicking on the specific server details. See the following video: https://i.gyazo.com/8be8ac04a739c847f801d24d76d10b35.mp4

By adding in those changes this issue doesn't persist as seen here: https://i.gyazo.com/269c9970bd3d5c6c7331c524d3611d85.mp4

My guess is that there's a callback that isn't getting called as a result of missing some HTML in the <td> tags, specifically data-status and data-pos. The re-ordering still works with these changes.

Zorrototo:
Origin of this PR: https://www.opengamepanel.org/forum/viewthread.php?thread_id=5359&rowstart=360&showlast=true#post_38276

Don't understand what his issue is, as we are multiple people to have tested with the SBS theme or official theme and we have no issue.

own3mall:
@xRJx Yeah, I can't tell what you're trying to show.  I don't see a problem?

xRJx:
When I click on either the address or owner fields to show the server info and start/stop buttons, that information won't show up. I'll use a different theme to demonstrate so that it'll be more clear. The only way that the information will show up is if you click on the server name. This is a small issue but I could imagine it causing someone confusion at first sight like it did for me.

This issue only occurs the first time you open the Game Monitor page, once that info pops up then it'll still be there when clicking on the address or owner fields until you revisit the page.

See here: https://i.gyazo.com/8a281fe73ff31bd2bd5a04f2176d6ce0.mp4

Zorrototo:
OK now I can reproduce with clearer explanation, if you click in the IP or OWNER box, yes, it doesn't load the content like it would if you click in the SERVER NAME box.

xRJx:
> OK now I can reproduce with clearer explanation, if you click in the IP or OWNER box, yes, it doesn't load the content like it would if you click in the SERVER NAME box.

Sorry, that was my bad. I didn't think to include that specific information until now.

own3mall:
Got it.  Sorry about that.  We didn't totally understand the issue being reported.

xRJx:
> Got it. Sorry about that. We didn't totally understand the issue being reported.

No worries, I'm just glad that you understand now.

Zorrototo:
On  a side note if you want to understand it all too, when he talked about the re-ordering, he was not talking about the sorting you do by clicking the column header, but the ordering you can manually do by clicking and dragging a cell to the position you want https://www.opengamepanel.org/forum/viewthread.php?thread_id=7048#post_37530 now everyone understands everything :100:

xRJx:
> On a side note if you want to understand it all too, when he talked about the re-ordering, he was not talking about the sorting you do by clicking the column header, but the ordering you can manually do by clicking and dragging the start of the row to the position you want https://www.opengamepanel.org/forum/viewthread.php?thread_id=7048#post_37530 now everyone understands everything

Oh cool, I didn't even realize it was possible to do that. Good to know!

Issue #481, opened 2020-04-22
PHP 7.3 - Server dashboard response for servers not empty


elitetheespeon:
I have recently upgraded from PHP 5.6 to PHP 7.3 and everything works except for the Dashboard page. When a server has any players, it creates the following PHP error:
```
PHP Fatal error:  Uncaught ArgumentCountError: Too few arguments to function print_player_list(), 3 passed in .../protocol/lgsl/LGSLMonitor.php on line 72 and exactly 4 expected in .../modules/dashboard/query_ref.php:25
Stack trace:
#0 ...protocol/lgsl/LGSLMonitor.php(72): print_player_list(Array, 3, 30)
#1 .../modules/dashboard/query_ref.php(76): require('...')
#2 .../includes/navig.php(106): exec_ogp_module()
#3 .../includes/navig.php(27): navigation()
#4 .../home.php(146): include('...')
#5 .../home.php(71): heading(true)
#6 {main}
  thrown in ...modules/dashboard/query_ref.php on line 25
```

own3mall:
Should be fixed now.  Thanks for reporting.

own3mall:
https://stackoverflow.com/questions/4697705/php-function-overloading

PR #484, opened 2020-04-24
Fix error for Dashboard

elitetheespeon:
After the fix for #481, I got error:
```
PHP Fatal error:  Uncaught Error: Call to undefined function print_player_list() in .../html/protocol/lgsl/LGSLMonitor.php:72
Stack trace:
#0 .../html/modules/dashboard/query_ref.php(76): require()
#1 .../html/includes/navig.php(106): exec_ogp_module()
#2 .../html/includes/navig.php(27): navigation()
#3 .../html/home.php(146): include('/home/vhosts/pa...')
#4 .../html/home.php(71): heading(true)
#5 {main}
  thrown in .../html/protocol/lgsl/LGSLMonitor.php on line 72
```
Looking at the code, I noticed it was still hitting the function `print_player_list()` inside of `protocol/lgsl/LGSLMonitor.php`, so I added a bypass if it comes in from the dashboard page. It now works again and pulls the server info.

### DIFF
--- a/protocol/lgsl/LGSLMonitor.php
+++ b/protocol/lgsl/LGSLMonitor.php
@@ -68,7 +68,8 @@ if($num_of_servers < $numberservers_to_skip_query)
 		$mapRaw = $data['s']['map'];
 		$address = $data['link'];
 		
-		if ( $data['s']['players'] > 0 )
+		if ( $data['s']['players'] > 0 && $_GET['m'] !== "dashboard" )
+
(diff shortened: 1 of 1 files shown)


own3mall:
Try the latest fix https://github.com/OpenGamePanel/OGP-Website/commit/d9b5c1dd5dd3aa0eaeb703f803d1bf57c2a4dcc1

Your solution works, but I don't want to filter by module when it comes to this code.  It's harmless to include the functions file here now that I've renamed the other function in use.

PR #496, opened 2020-06-03
Added bRawSockets support for ArkSE

MrCraigTunstall:
This is to better aid those whom have to use the bRawSockets command line in order for clients to join the server.

### DIFF
--- a/modules/config_games/server_configs/arkse_linux64.xml
+++ b/modules/config_games/server_configs/arkse_linux64.xml
@@ -5,7 +5,7 @@
   <installer>steamcmd</installer>
   <game_name>ARK: Survival Evolved</game_name>
   <server_exec_name>ShooterGameServer</server_exec_name>
-  <cli_template>%MAP%%IP%%PORT%%QUERY_PORT%%PLAYERS%%RCON%%CONTROL_PASSWORD%%PDS%%PDI%%PDD%%PUS%%PUI%%PUD%%ASDN%%POP%%POPI%%PTA%?listen %AMM% %CDO% %CID% %FACF% %NTFF% -server -log</cli_template>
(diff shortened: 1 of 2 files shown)


Zorrototo:
Are you sure about that? Isn't the command line parameter simply `bRawSockets` instead of `bRawSockets=true`? That's what I read on the official wiki, and if so then use a _checkbox_ instead of a _select_.

Also in that case, by using bRawSockets, wouldn't it mess up the way OGP works with this game server? would the Panel query work properly? (expecting game port + 19238 for query and not game port + 1 from what I read for bRawSockets)?

Did you test your XML?

MrCraigTunstall:
Yes, I tested the XML and all works correctly. I'll alter the XML later to just have ?bRawSockets than ?bRawSockets=True

MrCraigTunstall:
?bRawSockets and ?bRawSockets=True works by the looks of it.

Zorrototo:
and ?bRawSockets=False also works in the command line?

MrCraigTunstall:
Indeed as ?bRawSockets=False is the default. It's not required to put ?bRawSockets=False and can alter the XML to be an tickbox. Added it as True and False because some cases ?bRawSockets=True works whereas ?bRawSockets did not for a few server owners.

Zorrototo:
What I mean is does ?bRawSockets=False properly disable this feature because if the parameter ?bRawSockets=False ?bRawSockets=True and ?bRawSockets is interpreted the same in command line (because as I read in the wiki or on steam from devs, the parameter is just ?bRawSockets), then the tickbox is mandatory.

MrCraigTunstall:
Altered as requested, tested again to verify and works when adding that to the commandline. :)

Zorrototo:
OK, merging changes then.

Zorrototo:
For reference, using the parameter will need to have game port +1 UDP available, opened, and forwarded properly for this feature to work.

PR #505, opened 2020-07-07
[W.I.P] Basic support for RageMP

Zorrototo:
To be probably improved because it is :hankey:


Zorrototo:
Might be a good idea to add some kind of fail safe if the master list server is down. A better manipulation of the Json contained in the page could also yield more interesting info like the map, players/maxplayers. server name, but that is out of my capabilities (it would be simpler if the result was plain Json).

If someone has some time that would be cool to have some improvement on this mess I suggest to add :)

Zorrototo:
https://rage.mp/forums/topic/7519-querying-rage-mp-server/?tab=comments#comment-27287

Zorrototo:
Seems good to me with minimal functional support

Feel free to comment

PR #535, opened 2021-02-21
Add valheim xml

Bebiano86:


modules/config_games/server_configs/valheim_linux64.xml
```diff
@@ -0,0 +1,40 @@
+<game_config>
+ <game_key>valheim_linux</game_key>
+   <protocol>lgsl</protocol>
+  <lgsl_query_name>valheim</lgsl_query_name>
+ <installer>steamcmd</installer>
+ <game_name>Valheim</game_name>
+ <server_exec_name>valheim_server.x86_64</server_exec_name>
+ <cli_template>-name %HOSTNAME% %PORT% %HOME_PATH% %WORLD% %PASSWORD% -public= 1 </cli_template>
```
Zorrototo, line 8: I see blank spaces at the end of the line, is it correct? doesn't seem like it is to me without searching documentation

modules/config_games/server_configs/valheim_linux64.xml
```diff
@@ -0,0 +1,40 @@
...
+ <cli_params>
+  <cli_param options="s" cli_string="-savedir" id="HOME_PATH"/>
+  <cli_param id="PORT" cli_string="-port "  />
```
Zorrototo, line 11: If the port line needs a space, better use the proper option instead of manually adding blank space here

Bebiano86: I removed in the blanks was my mistake

Zorrototo:
Now that I looked at it more closely, I noticed multiple other mistakes.
For the %HOSTNAME% you didn't add the cli string with proper options (sq for space and quote I guess), you placed the cli string directly inside the cli template. I don't have the official list of options, but by default the value is added directly as is, without space, without quotes, if I would guess without looking into other XML, I would say

> s = adds a space before value
> q = adds quotes around value
> sq = adds a space and quotes the value

It is the same for all server parameters, I see you added spaces, instead of using proper OPTIONS in the XML. I looked in the WIKI it is not properly documented but take examples on other XML, try to make it correct by using the available options instead of manually adding spaces.

> Valid options are for the <options> element within the <param> element within the <server_params> element are:
> 
>         ns = no space between key and value
>         q = quotes wrapped around value after key (no space added)
>         s = space added after key before value (no quotes added)
>         anything else = space after key and quotes around the value 
>

Bebiano86:
thanks for the tips corrected.

ancientevil:
Does this XML work? When will it go into the repo?
3 million players :)

Bebiano86:
> Este XML funciona? Quando ele irá para o repo?
> 3 milhões de jogadores :)

we are making the last adjustments

ancientevil:
Great job! Will any1 contribute a win64 version aswell? @Bebiano86

Bebiano86:
> Great job! Will any1 contribute a win64 version aswell? @Bebiano86

yes later i can do that

Bebiano86:
done ;)

PR #550, opened 2021-03-25
Create valheimplus_linux64.xml

nuallan89:
The Plus version of the game . maybe some want just this one.. right ?


Zorrototo:
Without modifying the .sh script then, it will default to its content for server name port and all, and not take server startup parameters.

PR #556, opened 2021-07-06
Lang Update

OlliTheDarkness2020:
Create Lang German (Förmlich - Sie) and fill missing translations.

### DIFF
--- a/lang/German(Sie)/global.php
+++ b/lang/German(Sie)/global.php
@@ -1,0 +1,144 @@
+<?php
+/*
+ *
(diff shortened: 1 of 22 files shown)


Zorrototo:
Changes to language have to be made through Transifex https://www.transifex.com/opengamepanel/ogp/dashboard/ 
Unfortunately we can't (at least I can't) merge your changes directly this will be overwritten next synchronization of the Transifex files.

Zorrototo:
Maybe @rocco27 can see if he can force uploading the files, if the language is added.

PR #559, opened 2021-07-18
lack of words - addonsmanager.php

izmoxkawaii:


Zorrototo:
Sorry, but you need to help us translate OGP from Transifex, we can't accept an update from GitHub (unless it is on the Ebnglish language files) as it will be overwritten when the changes at Transifex will be pushed back.

https://www.transifex.com/opengamepanel/ogp/dashboard/

You can ask for access for your language there, then you can translate from the interface.

Sorry you wasted your time (but it is not all lost, you can now copy paste at Transifex).

PS: tell us if you asked for access to a language so I can make sure you're accepted as soon as possible.

PR #568, opened 2021-10-29
Fix cross-site-scripting vulnerability.

isaqueks:
Directly outputting raw values without `htmlspecialchars` can cause HTML to be rendered, which can be used to run malicious scripts codes to steal cookies for example. The script can be injected at login panel for example, typing HTML codes as Username. Then, when someone goes to `watch_logger`, the HTML codes are rendered.
This can be prevented by using `htmlspecialchars`.

### DIFF
--- a/modules/administration/watch_logger.php
+++ b/modules/administration/watch_logger.php
@@ -100,11 +100,14 @@ function exec_ogp_module() {
 		foreach($logs as $log)
 		{
 			$user = $db->getUserById($log['user_id']);
-			$when = $log['date'];
-			$who = $user['users_login'];
-			$where = $log['ip'];
-			$what = $log['message'];
-			$log_id = $log['log_id'];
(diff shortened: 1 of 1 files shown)


isaqueks:
Should these variables intentionally contain HTML codes? If not, I think it's a good practice to use `htmlspecialchars` because in the case that something goes wrong and HTML codes are assigned to the variables, those codes will be escaped.
If the variables **should** intentionally contain HTML, only escape the username.
PS: I am new to Open Game Panel.

own3mall:
This change isn't needed, as any sensitive variables should already be sanitized before being entered into the logger table.

I did a quick look through, and this was the only place that needed adjusting based on my initial look through:

https://github.com/OpenGamePanel/OGP-Website/pull/569

If there is an active vulnerability somewhere else for the watch logger, please provide detailed instructions to reproduce.

PR #581, opened 2022-05-11
OS icons replacement

Zorrototo:


Zorrototo:
To match font in menu, white cleaner icons

PR #583, opened 2022-06-08
Include Path fix for XXTEA

oNdsen:
This allows getting access to OGPRemoteLibrary Class from other Modules

### DIFF
--- a/includes/lib_remote.php
+++ b/includes/lib_remote.php
@@ -22,7 +22,7 @@
  *
  */
 
-require_once("Crypt/XXTEA.php");
+require_once(realpath(dirname(__FILE__)) . "/../Crypt/XXTEA.php");
 
 // Screen type for servers
 define("OGP_SCREEN_TYPE_HOME","HOME");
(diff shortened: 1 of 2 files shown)


own3mall:
Why did you change the path for XXTEA?  Where is this failing for you?  

In most of the code, lib_remote is loaded like this:

`require_once('includes/lib_remote.php');`

Thus, the active path is still currently the root OGP directory.  If that were not the case, all of the `require_once` calls would fail and prevent further execution of the code if the path weren't correct in the first place.

oNdsen:
My Goal are to implement additional Hooks in my Themes. To get Agent Informations like Crobjob i need to call to include the functions (i will not rewrite every function again...)

own3mall:
I still don't understand why you need this, but could you give me an example?

own3mall:
Ok, well I changed it here:

[Reference Path Directly from Script Location](https://github.com/OpenGamePanel/OGP-Website/pull/586)

PR #584, opened 2022-06-10
Update

Bebiano86:
new datePickerInput check method

### DIFF
--- a/js/modules/user_games.js
+++ b/js/modules/user_games.js
@@ -174,7 +174,7 @@ $(document).ready(function() {
 	});
 	
 	var datePickerInput = $("input[name=expiration_date]");
-	if(datePickerInput.length)
+	if(datePickerInput.length > 0)
 	{
 		var now_str = datePickerInput.attr('data-today').split(' '),
 			date = now_str[0].split('/'),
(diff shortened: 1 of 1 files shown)


Zorrototo:
Can you give more info about why you created this PR?

Bebiano86:
check never works as intended because the object "always exists", only with a length equal to its structure
in this case 0 (because there is no
hence the .length
which confirms whether the object is greater than 0

Zorrototo:
Is the issue happening only with this theme or is it a global issue wit hall themes? From my understanding it is a global issue but this is only with this theme that it seems to have happened (can't try right now).

Bebiano86:
it is a global problem the error is present in all themes.

oNdsen:
@Bebiano86 You check for boolean (datePickerInput.length). Please check if the Value is greather than 0 otherwise it will not work.
Check my example: https://github.com/oNdsen/OGP-Website/commit/bcaa19b518944a855ea152fe6144bac981c83ffd
@Zorrototo Yes it is a global Problem.

Bebiano86:
@oNdsen done

own3mall:
@oNdsen, this is not true.  Just checking for .length should be sufficient.  However, both work.

oNdsen:
Interessting, didnt worked with boolean check on my side. But maybe i just was too quick ;) Thanks for Check!

own3mall:
Fixed in https://github.com/OpenGamePanel/OGP-Website/pull/585

Thanks

PR #586, opened 2022-06-14
Reference Path Directly from Script Location

own3mall:


oNdsen:
Works, thanks for the Update!

PR #587, opened 2022-08-06
Update FiveM & RedM

HumanTree92:
- Update FiveM Max Slots to 2048
- Update RedM Max Slots to 1024
- Added Custom Field for sets sv_projectName
- Added Custom Field for sets sv_projectDesc
- Updated server.cfg to use FiveM's new server.cfg layout

The Max Slots for FiveM is 2048 & RedM is 1024 if you Subscribe to FiveM Patreon however default is set to 48 which is what FiveM now uses for servers that want the FREE OneSync. I also included their new Project Name & Project Desc into the Custom Fields to make it easier to Edit.

### DIFF
--- a/modules/config_games/server_configs/fivem_linux32.xml
+++ b/modules/config_games/server_configs/fivem_linux32.xml
@@ -5,8 +5,8 @@
   <game_name>FiveM</game_name>
   <server_exec_name>run.sh</server_exec_name>
   <cli_template>+exec server.cfg</cli_template>
-  <max_user_amount>32</max_user_amount>
-   <control_protocol>rcon</control_protocol>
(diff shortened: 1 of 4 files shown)


HumanTree92:
I have fully tested this on both FiveM & RedM. Everything works without any issues.

Zorrototo:
I will pass on to you the great power of trust, and will make myself a believer, that you did not send a rotten PR.

I don't own the game and can't test anything. Will assume it is correct and doesn't mess up an installation (from what I see it seems correct at least on the technical side of things, but can't keep up with recent server changes and all so I don't know anything about what these new and former settings do/did).

With the power of trust, I'm clicking...

[SQUASH AND MERGE]

click.

PR #588, opened 2022-08-22
Fix for Browser <form> closes and small Cleanups

oNdsen:
Since modern Browser render HTML realtime, a single <form> Tag will be closed automaticaly when not a close Tag will be written to Dom at the same Time.
Also it needs to have the <form> Tag within <tr> instead of single <td>, because there are additional Input Fields outside the <form> Tag - thats why i put the Content into an additional Table since there is the whole Default Theme styled with Tables.
In Cleanups i replaced .failure classes with Red Colors to avoid Error handling instead of colorizing and beautified a little the PHP Code.
No Errors with Revolution and Modern Themes like SBS and AdminLTE.
Reference: https://www.opengamepanel.org/forum/viewthread.php?thread_id=7898&rowstart=120#post_41080

### DIFF
--- a/modules/ftp/ftp_admin.php
+++ b/modules/ftp/ftp_admin.php
@@ -97,14 +97,16 @@ function exec_ogp_module()
 		$remote = new OGPRemoteLibrary($server_row['agent_ip'],$server_row['agent_port'],$server_row['encryption_key'],$server_row['timeout']);
 		$ftp_login = strip_real_escape_string($_POST['ftp_login']);
 		$settings = "";
+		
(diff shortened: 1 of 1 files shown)


Zorrototo:
While you're at it there is too much spaces before IPs in "Allowed local  IPs" https://github.com/OpenGamePanel/OGP-Website/pull/588/files#diff-0a643c40a028e5e4a14496a525266ac43f61a0f0ddfd9c93408cbbed991d1089R248
Not sure the implication of modifying that though.

oNdsen:
Same as before and it works

PR #596, opened 2023-04-26
Feature/php xmlrpc replacement

own3mall:
`php-xmlrpc` is unmaintained and appears to be unavailable as a package in newer operating systems via `apt` or `yum`.  I personally think this is another terrible decision by the php team, but whatever, it looks like we'll have to deal with it.

Source:

https://php.watch/versions/8.0/xmlrpc

Implemented Workarounds:

https://github.com/gggeek/phpxmlrpc
https://github.com/gggeek/polyfill-xmlrpc

Removed:

Requirements detecting php-xmlrpc during panel installation.  It will use native functionality now with the above included fixes in the source code.

Tested and Working in:

Latest Rocky Linux 9.x / RedHat 9.x
Ubuntu 14.04 (Oldest version OGP still supports - still have a few 14.04 legacy servers)

@DieFeM do you see any issues with this?


DieFeM:
Looks good. Thanks for taking care of it. It probably will raise some problem, but we will fix them. It's unavoidable.

Sh4d0wless:
Unable to download : https://github.com/OpenGamePanel/OGP-Website/commits/OGP-Website.atom
Unable to update: String could not be parsed as XML

PR #625, opened 2024-01-15
Update logger table to support IPv6 addresses by increasing IP column size to varchar(45)

moodejb123:


Zorrototo:
Related issue: https://opengamepanel.org/forum/viewthread.php?thread_id=8218

I wonder if we shouldn't add a routine/update to update current DB with new value too.

own3mall:
Completed in:  https://github.com/OpenGamePanel/OGP-Website/pull/626

moodejb123:
Thanks.

PR #630, opened 2024-02-23
Add chinese language

shinegod:
Add chinese language

### DIFF
--- a/lang/zh-cn/global.php
+++ b/lang/zh-cn/global.php
@@ -1,0 +1,144 @@
+<?php
+/*
+ *
+ * OGP - Open Game Panel
+ * Copyright (C) 2008 - 2018 The OGP Development Team
+ *
+ * http://www.opengamepanel.org/
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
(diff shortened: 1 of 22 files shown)


Zorrototo:
OGP translations will only be made through Transifex. 

https://app.transifex.com/opengamepanel/ogp/dashboard/

Chinese is not available there though. I'm trying to add it, then will try to see if I can upload your changes, we can the nsee how far the translation is and if we add it officially.

There are MANY language requests but adding a language nobody will support and maintain is kinda pointless. I guess Chinese may be good to add as the potential translators pool should be big so maybe we get one or two people interested in translating properly OGP.

Zorrototo:
I added Chinese to Transifex. Translation can be done there now.

Will see to add it to OGP officially later as there are things to do before it could be possible to send changes directly from Transifex to OGP's GitHub.

shinegod:
OK, I have applied to join and am waiting for review.

Zorrototo:
You have been added to the Chinese translator team ;)

Zorrototo:
OK, @own3mall did what was required for the sync to happen, there is now a new Chinese(China) language available in OGP, and translation to Panel and modules should be synchronized with Transifex now.

PR #632, opened 2024-03-28
Update csgo_linux.xml

Nairdaa:
"The CS2 dedicated server and game client have been merged together (appid 730). Previously the CSGO client (appid 730) and dedicated server (appid 740) were separate."


Zorrototo:
There is a discussion about this topic here https://opengamepanel.org/forum/viewthread.php?thread_id=8064

Current situation is not satisfactory regarding how to handle the CSGO/CS2 server/game files (currently CS 2 requires the game files to be downloaded and some intervention to have it to work).

PR #640, opened 2025-03-10
Update hldm_linux32.xml

duncosta:


Zorrototo:
Is the current XML not working, are these changes required? Some context?

duncosta:
Thix xml it is for half life servers not for counter strike 1.6 .

own3mall:
Last I checked, the files get downloaded into the cstrike folder... so unless something has changed... I guess I'll have to recheck when I get a chance.

duncosta:
Screenshot_1

own3mall:
@duncosta , thanks you're right.

PR #641, opened 2025-03-10
Update hldm_win32.xml

duncosta:


duncosta:
Thix xml it is for half life servers not for counter strike 1.6 .

own3mall:
Last I checked, the files get downloaded into the cstrike folder... so unless something has changed... I guess I'll have to recheck when I get a chance.

own3mall:
@duncosta , thanks you're right.

PR #643, opened 2025-03-31
CS2 on Linux 64-bit systems

astral123457:
A PHP script potentially related to managing or implementing functionalities for CS2, including server communication and game settings.
An XML configuration file for defining server settings and operational parameters for CS2 on Linux 64-bit systems.


own3mall:
@astral123457 , have you tested this?  Also, how does it install the game?  I don't see steamcmd being used with app 730.

Zorrototo:
The state of CS2 is still bad as far as I know. You need to install the game, do some manual intervention in the files, and then you can start a server. You can follow along this thread https://opengamepanel.org/forum/viewthread.php?thread_id=8064 where work on a XML was done, I personally advised against official support for it as long as they don't have a proper server AppID for the server and proper support for it. If a team member is OK with integration of a XML for CS2 it's OK but on my side, no proper support from Valve for server = no possible proper support on our side for a CS2 server in OGP.

PR #652, opened 2025-10-07
CSS Fixes - Better Than Before

own3mall:


own3mall:
Better for older browsers that don't support the newest CSS standards.

PR #666, opened 2026-03-24
Fix directory traversal vulnerability and double URL decode in litefm

NiceFeatures:
This Pull Request addresses critical bugs within the OpenGamePanel File Manager (litefm) module, improving its security and fixing file resolution issues.

1. File Manager Injection / Path Traversal Validation (fm_dir.php)
Batch operations (such as rename, remove, move, copy, compress, uncompress, send_by_email and secure_file) in fm_dir.php were fetching and executing file commands directly from the $_SESSION['fm_files_'] array without adequate strict path validation.
An issue occurred when a malicious or malformed filename, such as ;test.txt, was manipulated. Because the remote server execution splits arguments using ;, it caused the original path to break prematurely. For example, renaming ;test.txt would send rename("/path/", "test.txt"), potentially erasing or altering the entire parent directory.
Fix: A strict Regular Expression guard (preg_match("/\/\.\.\/|\||;/", $item_path)) has been applied inside the loops of all batch operations. It immediately aborts the process and throws an unallowed_char error if standard injection characters like ; or | are detected in the session paths.

2. File Name Double-Decode Bug (litefm.php)

Filenames containing special URL-encoded characters (most notably the + sign) were failing to be found or modified by the File Manager.
The original code applied a urldecode() call when obtaining the filename: $fileName = !empty($_POST['name']) ? urldecode($_POST['name']) : urldecode($_GET['name']);
Since PHP's web server engine natively URL-decodes variables inside $_GET and $_POST superglobals automatically, this resulted in a destructive Double Decode. A valid + character in a remote filename would get decoded a second time into a blank space ( ), which inevitably caused "File Not Found" errors or phantom file creations.
Fix: Removed the redundant urldecode() wrapper around $_POST['name'] and $_GET['name']. The application now relies on PHP's correct built-in parameter decoding.
Testing performed
•	Successfully validated PHP syntax for both patches using php -l.
•	Verified file handling and agent side script logic to confirm that correct behavior is preserved.

### DIFF
--- a/includes/helpers.php
+++ b/includes/helpers.php
@@ -495,4 +495,34 @@ function getOGPLangConstantsJSON(){
 	
 	return false;
 }
+
+/**
+ * Validates a file path for security before passing to RPC/Shell.
+ *
+ * @param string $path The full path to validate.
+ * @return bool True if safe, False if it contains dangerous characters or traversal.
+ */
+function validate_path($path) {
+    // 1. Block any instance of ".." (Parent Directory Traversal)
+    if (strpos($path, '..') !== false) {
+        return false;
+    }
+
+    // 2. Block "." if it is the entire filename or a standalone path segment
+    // This prevents operations on the "current directory" itself.
+    $parts = explode('/', $path);
+    foreach ($parts as $part) {
+        if ($part === '.') {
+            return false;
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


own3mall:
@NiceFeatures Thanks!

